PR #2, opened 2017-01-04
RadekP2 tasks

radekpetruska:
- Modify List.jsx to render "todo" list.
- Add new components that allow to add/update/delete list items.


radekpetruska:
All bullet points should be reflected.

Suzii:
Good job with the changes :) 
I don't think it is a good idea to pass only test as a string to ListItemContainer, though. Object with text property was better as it is easy to extend with more properties (as was mentioned in the Task 1 description).
If you decide to change it, have a look at the React.PropTypes.shape({ ... }) -- cool stuff ;)

Another thing considering Draft's coding style - we use static property declared inside of the class for defining propTypes, instead of specifying them outside, e.g:
```
class SomeComponent extends React.Component {
  static propTypes = {
      projectId: PropTypes.string.isRequired,
  };

  constructor() { ... }

  // other methods
}
```
But of course, your solution is equivalent and it is just matter of internal team guidelines which one should be prefered. :)

radekpetruska:
Let's do it in Draft way.

Suzii:
Great :) (Y) 
Note: I will leave this pull request open so that you can submit other tasks as well

radekpetruska:
That should be all for the first attempt to implement task #2.
I've kept states in child components since I didn't see any sense to move them to the root. Just tell me if I overlooked something. Specifically:
- NewItem.jsx - current text in text box
- ListItemEditableText.jsx - current text in text box
- ListItemContainer - currently rendered mode

Suzii:
As for the "all state in root component"... Generally it's better to have all the state at one place and propagate down via props (or once you have flux/redix it's even more starightforward). In case of this task, I would surely leave NewItem.jsx as it is right now. The question is whether or not to move the edited value up. I am not saying your solution is bad, you can leave it as is, since it does not really make sense in this small example. But if you had an item with more properties that could be changed or its values could be chenged "from outside", or its current value changes something else in the app -- then you should consider moving it to root (or edux store) :)

radekpetruska:
Task \#3:
Application state is now completely stored in Redux store => your previous suggestions were reflected in code:
- _updateState() is gone
- Both edited value and rendered mode is driven by props retrieved from Redux store

onboarding-task/src/actions/actionTypes.js
```diff
@@ -0,0 +1,10 @@
+const NEWITEM_UPDATE = 'NEWITEM_UPDATE';
+
+const ITEM_INSERT = 'ITEM_INSERT';
+const ITEM_CHECKOUT = 'ITEM_CHECKOUT';
+const ITEM_UPDATE = 'ITEM_UPDATE';
+const ITEM_REVERT = 'ITEM_REVERT';
+const ITEM_CHECKIN = 'ITEM_CHECKIN';
+const ITEM_DELETE = 'ITEM_DELETE';
+
+export {NEWITEM_UPDATE, ITEM_INSERT, ITEM_CHECKOUT, ITEM_UPDATE, ITEM_REVERT, ITEM_CHECKIN, ITEM_DELETE};
```
Suzii: You can use export default if you grouped the constants into an object. Alternatively use export const for each constant you want to export. But I would say one default export is preffered :)

onboarding-task/src/reducers/itemActionsReducer.js
```diff
@@ -0,0 +1,109 @@
+import Immutable from 'immutable';
+import {ITEM_INSERT, ITEM_CHECKOUT, ITEM_REVERT, ITEM_CHECKIN, ITEM_UPDATE, ITEM_DELETE} from '../actions/actionTypes';
+import createGuid from '../utils/guidGenerator';
+import {ID, TEXT, IS_CHECKED_OUT, ORIGINAL_TEXT} from '../descriptors/itemProperties';
+
+const insertItem = (state, itemText) => {
+  const newItem = Immutable.Map({
+    [TEXT]: itemText,
+    [ID]: createGuid(),
```
Suzii: You should never generate stuff in reducers. Reducer MUST be a pure function without any side effects and therefore must never call API or do things such Math.Random() or new Date(). Move either item as a whole to action creator or at least pass newlu generated ID with text to be added as payload of the action.

onboarding-task/src/reducers/itemActionsReducer.js
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  const updatedItems = state
+    .update(itemId, item => item.set(TEXT, newText));
+
+  return updatedItems;
+};
+
+const itemActionsReducer = (previousStoreState = Immutable.Map(), action) => {
+  switch (action.type) {
```
Suzii: I would suggest you try appling reducer composition even on this level. State is a Map and your payload should always contain ID of an item (key in Map). Therefore, you can simplyfy other methods to take only the relevent item istead of whole Map where applicable (-> I think only insert and delete would have to be handled separately).

onboarding-task/src/reducers/combinedReducer.js
```diff
@@ -0,0 +1,10 @@
+import {combineReducers} from 'redux';
+import ItemActionsReducer from './itemActionsReducer';
```
Suzii: Why UpperCamelCase?

onboarding-task/src/components/NewItem.jsx
```diff
@@ -1,43 +1,41 @@
...
-          <button className="btn btn-default" onClick={this._insertItem}>Add</button>
+          <button className="btn btn-default" onClick={this.props.insertItem}>Add</button>
         </div>
       </div>
     );
   }
 }
 
-export default NewItem;
+const mapStateToProps = (dispatch, ownProps) => {
```
Suzii: I guess this should be renamed to 'mapDispatchToProps'.

onboarding-task/src/components/NewItem.jsx
```diff
@@ -1,43 +1,41 @@
 import React from 'react';
+import {connect} from 'react-redux';
+import insertItem from '../actions/insertItem';
+import updateNewItem from '../actions/updateNewItem';
 
 class NewItem extends React.Component {
   static propTypes = {
-    onInsert: React.PropTypes.func.isRequired
+    text: React.PropTypes.string.isRequired
```
Suzii: Why not updateText() and insertItem() functions as well?

onboarding-task/src/components/ListItemEditableText.jsx
```diff
@@ -1,46 +1,50 @@
...
+import {connect} from 'react-redux';
 import ImmutablePropTypes from 'react-immutable-proptypes';
+import deleteItem from '../actions/deleteItem';
+import updateItem from '../actions/updateItem';
+import checkItemIn from '../actions/checkItemIn';
+import revertItem from '../actions/revertItem';
+import {TEXT} from '../descriptors/itemProperties';
 
 class ListItemEditableText extends React.Component {
   static propTypes = {
```
Suzii: Specify callback functions which are passed as props as well.

onboarding-task/src/components/ListItemContainer.jsx
```diff
@@ -1,52 +1,27 @@
...
 import ReadOnlyText from './ListItemReadOnlyText';
 import EditableText from './ListItemEditableText';
-
-const ReadOnlyMode = Symbol('readonly');
-const EditMode = Symbol('edit');
+import checkItemOut from '../actions/checkItemOut';
+import {TEXT, IS_CHECKED_OUT} from '../descriptors/itemProperties';
 
 class ListItemContainer extends React.Component {
   static propTypes = {
```
Suzii: Add all calback functions as well.

radekpetruska:
ad) descriptors: I haven't seen this before but I just wanted to have some constants to avoid typos since item properties are used on a lot of places + it would be easier to rename properties if needed.

onboarding-task/src/App.jsx
```diff
@@ -1,35 +1,45 @@
...
+import createLogger from 'redux-logger';
+import reducer from './reducers/combinedReducer';
 import List from './components/List';
 
 class App extends Component {
   render() {
+    const logger = createLogger();
+    const storeEnhancers = applyMiddleware(logger);
+    const store = createStore(reducer, storeEnhancers);
+
```
Suzii: It is always good to avoid instantiation of objects especially as important as store object in `render()` method. Render is called many times in component lifecycle and therefore new instance will be created every time. You can argue that `App.jsx` should be rendered exactly once but it is a good practice to place this stuff outside anyway. Above class definition after all imports will do :)

onboarding-task/src/actions/actionTypes.js
```diff
@@ -0,0 +1,11 @@
+class actionTypes {
```
Suzii: Why creating a class where simple object literal suffices?

onboarding-task/src/components/ListItemEditableText.jsx
```diff
@@ -0,0 +1,54 @@
...
+        <button className="btn btn-primary" onClick={this.props.confirmChanges}>Save</button>
+        <button className="btn btn-default" onClick={this.props.revertChanges}>Cancel</button>
+        <button className="btn btn-danger" onClick={this.props.deleteItem}>Delete</button>
+      </div>
+    );
+  }
+}
+
+const mapDispatchToProps = (dispatch, ownProps) => {
+  return {
```
Suzii: I am not really sure whether this approach is better to sending parameters explicitly from Component. I mean, component should be as much reusable as possible on the other hand, why should redux container be responsible for extracting input value from event object? Personally, I would not use own props at all but that's just my opinion, you don't have to agree and I believe both approaches have their pros and cons.

onboarding-task/src/components/ListItemEditableText.jsx
```diff
@@ -0,0 +1,54 @@
...
+    updateText: event => {
+      const action = updateItem(ownProps.item, event.target.value);
+      dispatch(action);
+    },
+    confirmChanges: text => {
```
Suzii: unused parameter?

onboarding-task/src/reducers/itemActionsReducer.js
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  return itemToCheckIn
+    .set(itemProperties.IS_CHECKED_OUT, false)
+    .set(itemProperties.ORIGINAL_TEXT, itemToCheckIn.get(itemProperties.TEXT));
+};
+
+const updateItem = (itemToUpdate, newText) => {
+  if (!itemToUpdate.get(itemProperties.IS_CHECKED_OUT)) {
+    throw new Error('Cannot update checked in item.');
```
Suzii: Not really sure if throwing in reducer is a good idea. See: https://github.com/reactjs/redux/issues/1960... you can always move this check to actionCreator. I do not have much experience with this, but it feels bad and seems like a side effect from reducer. Let me know your opinion.

radekpetruska:
1. I thought so but I was not sure since the render was called only once :smile: 
2. Because C# :smile: 
3. Since I have no previous experience, I changed it so it does not use ownProps. I liked the previous solution just because it was shorter in jsx
4. Resolved in 3)
5. I previously put it here since there is nothing that can stop me from sending actions created by hand (with checked in item) + it helped me to track bugs in unfinished implementation. But since reducers are not expected to throw errors I moved the check to action creator. There is one little change though: according to the article, it's not good to throw an error anyway, so it only log message to console and continue in standard flow.

onboarding-task/test/reducers/insertFormActionsReducer.test.js
```diff
@@ -0,0 +1,27 @@
...
+  it('Updates item text', () => {
+    const originalState = Immutable.Map({
+      text: 'pasted text'
+    });
+
+    const action = updateNewItem('updated text');
+    const state = insertFormActionsReducer(originalState, action);
+
+    expect(originalState.get(properties.TEXT)).toBe('pasted text');
+    expect(state.get(properties.TEXT)).toBe('updated text');
```
Suzii: You may also be interested in checking that no other properties of the item were changed.

onboarding-task/test/testUtils/initialStateGenerator.js
```diff
@@ -0,0 +1,17 @@
+import Immutable from 'Immutable';
+import itemGenerator from './itemGenerator';
+import checkItem from './checkItem';
+
+const defaultItemId = '00-00';
+const defaultItemText = 'default item';
+
+export const createInitialState = () => {
```
Suzii: Exporting one default object at the end of the file would be preferred.

onboarding-task/test/testUtils/checkItem.js
```diff
@@ -0,0 +1,10 @@
+import itemProperties from '../../src/descriptors/itemProperties';
+
+const checkItem = (itemToCheck, text, id, isCheckedOut, originalText) => {
```
Suzii: This is a bit of magic, too many parameters where the order is super-important. It might be nice to create something more complex and dumb-proof, like creating closure with `item` and returning an object with `ceckXY(xy)` functions wich would return the outer function and might be called in fluent syntax. In case of a failing check, an error could be thrown? I cannot think of anything simpler right now, we can talk in person about other possible options :)

PR #3, opened 2017-01-13
PetrSvirak - Task 1-3 + some extra buttons

PetrSvirak:
* functionality based on ``assignment.gif``
* almost missed the "random" identifier
* hopefully did not miss (too much of) other things


PetrSvirak:
The conventions we discussed should be all in; RfR.

PetrSvirak:
One note to ``ImmutablePropTypes.recordOf``: it seems that it only handles properties. It was somewhat wrong yet reassuring to have a property type requiring a ``set`` method to be present in them ``Item`` object…

PetrSvirak:
Thanks.

I have hopefully implemented all your suggestions and fixed all problems you pointed out (and few details I previously oversaw). 

You are right that extensibility and independence of ``EditItem`` do not make much sense here. 

Movement of ``<li>``s to ``List`` component also makes great sense, yet I noticed that "clickable" area is somewhat limited by the styling. This behavior is present from the very beginning as the ``onClick`` handler was always bound to the item's text rather than ``<li>`` element itself, however, it is now harder to mitigate such unpleasant UX. Would you, in case it was a problem, suggest moving ``<li>`` element down to the ``ExistingItem`` component?

PS: Lowercased methods gonna hurt for sometime :angel:

PetrSvirak:
Fixed :D

onboarding-task/src/containers/NewItem.jsx
```diff
@@ -0,0 +1,18 @@
+import * as ReactRedux from 'react-redux';
+import NewItemComponent from '../components/NewItem';
+import addItemAction from '../actions/addItem';
+
+const mapStateToProps = () => ({});
```
PetrSvirak: But, but, it's the _first_ argument. It says "mapStateToProps() in Connect(NewItem) must return a plain object. Instead received ``null`` (or ``undefined``). It seems to be possible to pass ``undefined`` directly to the _connect_ method though :)

onboarding-task/src/reducers/items.js
```diff
@@ -0,0 +1,27 @@
+import * as Immutable from 'immutable';
+import Item from '../models/Item';
+import { ITEM_ADD, ITEM_DELETE, ITEM_UPDATE_DESCRIPTION } from '../actions/actionTypes';
+
+function items(state = Immutable.OrderedMap(), action) {
+  switch (action.type) {
+    case ITEM_ADD: {
+      const newItem = new Item(action.description);
```
PetrSvirak: Yes, I'm completely sure. First of all, I wrote both. Also, Item creates an instance of an immutable record. However, if You suggest that reducer should not be aware of Item and ensure it does not incorporate any unintended logic or side-effects, new item creation can (~ will) be moved to the action builder.

PetrSvirak:
As for the action type ``ITEM_UPDATE_IS_EDITED``: it is used for both start of the item edition and its cancellation (cancel button), while ``ITEM_UPDATE_DESCRIPTION`` is triggered solely by description change (update button).

onboarding-task/src/index.js
```diff
@@ -1,10 +1,23 @@
...
+import { Provider as ReduxProvider } from 'react-redux';
+import createLogger from 'redux-logger';
+import appReducer from './reducers/app';
+import { seedStore } from './utils/store';
+import App from './App';
 import 'bootstrap/dist/css/bootstrap.css';
 import './index.css';
 
+const logger = createLogger();
+const store = Redux.createStore(
```
PetrSvirak: They do now. I initially fixed it to another branch 😇

PetrSvirak:
I have added the button we discussed earlier and one for restoring original description.

What I wonder is whether action creators themselves should or should not create immutable objects. What is your opinion on this?

onboarding-task/src/models/EditedItem.js
```diff
@@ -0,0 +1,11 @@
+import { Record } from 'immutable';
+
+const defaultNewItem = {
```
Suzii: Name does not make much sense in this context, copy-paste?

onboarding-task/src/reducers/editedItems.js
```diff
@@ -1,20 +1,32 @@
...
+        .toMap();
+    }
+    case (ITEM_STORE_EDITED_DESCRIPTION): {
+      // Stores edition description for any currently edited item
+      return state.set(payload.id, payload.editedItem);
+    }
     case (ITEM_UPDATE_DESCRIPTION): {
       // Renders item without edit mode after description update
-      return state.delete(action.id);
+      return state.delete(payload.id);
```
Suzii: You can merge the statement body with the two follow up cases, as they are the same.

onboarding-task/src/models/EditedItem.js
```diff
@@ -0,0 +1,11 @@
...
+  description: '',
+  isOriginal: true,
+};
```
Suzii: Store is as well?

onboarding-task/src/utils/item.js
```diff
@@ -0,0 +1,10 @@
+import * as Immutable from 'immutable';
+import { isNotEmpty } from './text';
+
+const getStorableEditedItems = editedItems => Immutable
+  .Seq(editedItems)
+  .filter(item => !item.isOriginal)
+  .filter(item => isNotEmpty(item.description))
+  .toMap();
+
+export { getStorableEditedItems };
```
Suzii: can be made default

onboarding-task/src/utils/store.js
```diff
@@ -14,7 +14,7 @@ const getStaticItemsDictionary = () => getStaticItems()
 const seedStore = () => {
   return {
     items: new Immutable.OrderedMap(getStaticItemsDictionary()),
-    editedItems: new Immutable.Set(),
+    editedItems: undefined,
   };
 };
```
Suzii: export can be made default

onboarding-task/src/utils/text.js
```diff
@@ -1,3 +1,3 @@
-const isStorable = text => (text || '').trim().length > 0;
+const isNotEmpty = text => (text || '').trim().length > 0;
 
-export { isStorable };
+export { isNotEmpty };
```
Suzii: can be made default

onboarding-task/test/reducers/editedItems.test.js
```diff
@@ -1,48 +1,113 @@
...
-    const item = new Item('test description');
-    const action = updateItemIsEdited(item.id, true);
+  it('cancel edition action deletes item if present', () => {
+    const itemId = 'baubaubauba';
+    const editedItem = new EditedItem();
+
+    const action = cancelItemEdition(itemId);
+    const currentState = new Immutable
+      .Map()
+      .set(itemId, editedItem);
```
Suzii: You can use `new Immutable.Map({ idemId: editedItem })`. Your solution creates unnecessary intermediate objects.

onboarding-task/src/components/UpdateAllEditedItems.jsx
```diff
@@ -0,0 +1,39 @@
...
+        description: PropTypes.string,
+        isEdited: PropTypes.bool,
+      }),
+      PropTypes.string,
+    ),
+    storableItemsCount: PropTypes.number.isRequired,
+    updateAllItems: PropTypes.func.isRequired,
+  };
+
+  _updateAllItems = () => this.props.updateAllItems(this.props.storableItems);
```
Suzii: You can use ownProps in the container to pass the `storableItems` to action as you did in other containers.

onboarding-task/src/components/UpdateAllEditedItems.jsx
```diff
@@ -0,0 +1,39 @@
...
+class UpdateAllEditedItems extends Component {
+  static propTypes = {
+    storableItems: ImmutablePropTypes.mapOf(
+      ImmutablePropTypes.recordOf({
+        description: PropTypes.string,
+        isEdited: PropTypes.bool,
+      }),
+      PropTypes.string,
+    ),
+    storableItemsCount: PropTypes.number.isRequired,
```
Suzii: It is recommended not to send props that can be computed from other props.

onboarding-task/src/models/EditedItem.js
PetrSvirak: https://briankoberlein.com/wp-content/uploads/copypasta.jpg :innocent:

onboarding-task/src/models/EditedItem.js
PetrSvirak: Took me a while to get your meaning here, but: nope.

onboarding-task/src/components/UpdateAllEditedItems.jsx
PetrSvirak: I don't believe so. Function ``mapDispatchToProps`` seems not to have access to the values set in ``mapStateToProps``, so there are no ``storableItems``. We can check this one out in person. I might miss something here, howver, ``ownProps`` cause component to misbehave…

onboarding-task/src/components/UpdateAllEditedItems.jsx
PetrSvirak: I actually did this, so component does not need anything about the items themselves. I would ideally not sent ``storableItems`` at all (see below), but, in this case, I also enjoy the fact that there is no logic in the component itself and all computations and transformations are done in "its" container…

PetrSvirak:
Here is my initial motivation connected with ``item`` and ``store`` and ``text`` modules in the ``utils`` folder: There is definitely just one method in each module, true. However, if there were multiple methods, there would be none that I would mark as "default", having this in mind, I purposely omit ``default export``, so other method's addition would not result in refactoring of the inclusion of existing method(s). 
I'm fully, as always, opened to discussion, so we can clear this up (for me) :)

PetrSvirak:
And thank you for your kind words :)

Suzii:
As for the `default export` in `utils/` folder. You used `export { someUtilFunction }` -- this means you are exporting an `object` with `someUtilFunction`. Therefore you can extend this object with more functions and still just one object will be exported and imports independent files stay exactly the same :) Problems you describe would come if you exported just one function.

onboarding-task/src/components/UpdateAllEditedItems.jsx
Suzii: Sorry, I missed that one :)

onboarding-task/src/components/UpdateAllEditedItems.jsx
Suzii: Sorry, I missed that one :)

PetrSvirak:
Base on your comment and our personal discussion and my further experience: I do not believe that ``export { … };`` and ``export default { … }`` are same thing. Please, see [this SO (answer)](http://stackoverflow.com/a/36796281/1138663) and [this SO (answer)](http://stackoverflow.com/a/34668968/1138663).

While my original code is actually just grouped named export and should be equivalent to, let's say, this:
````
export Func1;
export Func2;
````

The solution with default one should be equivalent to:
````
const group = ({ Func1, Func2 });
export (default) group;
````

Each approach results in different behavior when importing from those utility modules. In the first case:
````
import { Func1 } from 'utils/…';
 …
Func1();
// Func2(); is not available and calling it will fail
````
While, in the second case, only following usage seems to work:
````
import utils from 'utils/…';
 …
utils.Func1();
utils.Func2();
````

I might get something wrong, sorry if it is the case (again 😇); however, I personally prefer the approach I've chosen in the first place (no strong opinion on this one, though). Referring back the first [SO answer](http://stackoverflow.com/questions/36795819/when-should-i-use-curly-braces-for-es6-import/36796281) I mentioned earlier (its end, actually), I believe that none of ``utils`` modules actually have any normal functionality representing them and thus, IMHO, only named exports should (probably) be used.

Suzii:
Okay, sorry for the confusion. I have clearly missunderstood what `export { someUtilFunction }` does. It does not export an object with that property, instead it is only a shorthand syntax for renaming your exports. I guess all the other review comments were reflected, so I'm merging the pull request :)

PR #7, opened 2017-02-01
Finished task 2

ivanjrn:
- Lists's state now uses immutable map
- AddItem refactored to NewItem as suggested by Zuzka in my previous pull request
- handleEvent function names in List component refactored to onEvent
- editable property in List's state refactored to isEdited

### DIFF
--- a/onboarding-task/package.json
+++ b/onboarding-task/package.json
@@ -54,6 +54,7 @@
     "@types/react": "^15.0.0",
     "@types/react-dom": "^0.14.20",
     "bootstrap": "^3.3.7",
+    "immutable": "^3.8.1",
     "react": "^15.4.2",
     "react-dom": "^15.4.2"
   },
--- a/onboarding-task/src/components/List.jsx
+++ b/onboarding-task/src/components/List.jsx
@@ -50,7 +50,7 @@ class List extends Component {
   }
 
   _saveItem(guid, text) {
-    const items = this.state.items.updateIn([guid, 'text'], val => text).updateIn([guid, 'isEdited'], val => false);
+    const items = this.state.items.updateIn([guid, 'text'], () => text).updateIn([guid, 'isEdited'], () => false);
     this.setState({ items });
   }
 
--- a/onboarding-task/src/components/List.jsx
+++ b/onboarding-task/src/components/List.jsx
@@ -11,7 +11,7 @@ class List extends Component {
 
   constructor() {
     super();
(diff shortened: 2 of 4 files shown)


Suzii:
Very nice and clean solution, I like it very much :) I have only some minor comments:
- It looks like you don't have `Immutable` listed in you dependencies in `package.json`. This typically happen when you use command `npm install immutable` instead of `npm install --save immutable` (or any other library). Be very carefull with this. First command only downloads the library to `node_modules` folder, whereas second command (with `--save` switch) also adds the dependency to `package.json`. This way you share the dependencies with your team. When your teammate runs `npm install`, only dependencies listed in `packages.json` will be downloaded :)
- Have a look at `immutable-proptypes` library and try to use it. Right now you have ` item: PropTypes.object.isRequired,` in `PropTypes` which is not really the case since `item` is an `Immutable.Map`.
- `ListItemEditable.jsx`: rename the `_onClick()` method. (I suppose it should have been something like `_onCancel()` in the first place.
- `List.jsx`: I like how you extracted `getInitialState()` into a separate method. Two things here, though:
  - Add the underscore since in ES6 I think it is not a React lifecycle method (correct me if I am wrong).
  - The methods says it is a `get..` but does not return anything. Therefore, I would suggest refactoring the constructor to something like `this.state = this._getInitialState()`. And of course the method would return the Map with items.

Suzii:
As for the initial problem with pull requests, it turned out that when you submit a PR from your `develop` branch, all the commits you will add after PR is created, are automatically added to the PR if it is not closed yet. As a result, we decided to change the workflow a bit. Please submit the next PR (with Redux implementation) from the `features/task-3` branch, so that it is not mixed in this review. Also, please do not mergre your task 3 into your `develop` branch before PR is approved :) Any changes you do to task 2 based on feedback from this review shall be commited to task 2 branch. Thanks in advance.

ivanjrn:
- Added Immutable to package.json
- Introduced immutable-proptypes in ListItemStatic and ListItemEditable
- Refactored _onClick() function to _onCancel()
- Refactored getInitialState() function as suggested

Suzii:
Nice, thanks for reflecting the comments :) .. By the way, by immutable-proptypes I meant https://www.npmjs.com/package/react-immutable-proptypes, which allows you to check the contents of the Map you get as a prop :) I am closing this pull request. If you want, you can use the immutable-proptypes in the next task :)

PR #15, opened 2017-02-14
Features/task 2

JiriLojda:


onboarding-task/src/components/ListItem.jsx
```diff
@@ -13,11 +18,21 @@ class ListItem extends Component {
   constructor(props) {
     super(props);
     this._onSubmit = this._onSubmit.bind(this);
+    this._switchFormDisplayed = this._switchFormDisplayed.bind(this);
```
Suzii: Keep the style consistent when naming the methods. Now you heve `on` prefix in some and other times you don't.

onboarding-task/src/components/List.jsx
```diff
@@ -44,7 +41,7 @@ class List extends Component {
 
   _switchFormDisplayedOnId(id) {
     const item = this.state.items.get(id);
-    const newItems = this.state.items.set(id, { ...item, formDisplayed: !item.formDisplayed });
+    const newItems = this.state.items.set(id, { id, text: item.text, formDisplayed: !item.formDisplayed });
```
Suzii: Aren't you mixing plain JS objects and Immutable.Record here?

onboarding-task/src/components/List.jsx
```diff
@@ -19,19 +20,15 @@ class List extends Component {
...
   }
 
-  _createFunctionWithBoundId(id, func) {
-    return func.bind(this, id);
-  }
-
   _onListItemAdd(text) {
     const id = guid();
-    const newItems = this.state.items.set(id, { id, text, formDisplayed: false });
+    const newItems = this.state.items.set(id, new ItemRecord({ id, text, formDisplayed: false }));
```
Suzii: Please, refactor the new item creation to separate variable for better readability :) Do same also for other methods in this file.

onboarding-task/src/components/List.jsx
```diff
@@ -19,19 +20,15 @@ class List extends Component {
...
     const newItemsOrder = this.state.itemsOrder.push(id);
     this.setState({ items: newItems, itemsOrder: newItemsOrder });
   }
 
   _onListItemSubmit(id, text) {
-    const newItems = this.state.items.set(id, { ...this.state.items.get(id), formDisplayed: false, text });
+    const newItems = this.state.items.set(id, new ItemRecord({ id, formDisplayed: false, text }));
```
Suzii: If you had more properties on the `Record`, not just those three, it would be a lot better to get the old item, change the two properties (`text` and `formDisplayed`) and set `updatedItem` to the list.

onboarding-task/src/components/List.jsx
```diff
@@ -19,19 +20,15 @@ class List extends Component {
...
-
   _onListItemAdd(text) {
     const id = guid();
-    const newItems = this.state.items.set(id, { id, text, formDisplayed: false });
+    const newItems = this.state.items.set(id, new ItemRecord({ id, text, formDisplayed: false }));
     const newItemsOrder = this.state.itemsOrder.push(id);
     this.setState({ items: newItems, itemsOrder: newItemsOrder });
   }
 
   _onListItemSubmit(id, text) {
```
Suzii: `Submit` here is a bit missleading, isn't it more of an `update`, or `textUpdate`?

Suzii:
I was told my reviews are too lenient, so I have some more suggestions for you :D 

`ListItem.jsx`:
 - It would make more sense to move the `<li ...>` element up to the `List.jsx` since that component is responsible for rendering the list as a whole I guess.
 - I would maybe like to see the "edit" and "display"  logic of the items as separate components, does it make sense for you?

And I will talk to Peter and probably update the coding rules about naming the methods in components later today. As for the props callbacks deffinitely stick to `on` prefix as you did, but the method names look a bit cumbersome here. We can discuss it in person :)

PR #17, opened 2017-02-16
Feature/task 1

davidcechak:
Task 1 refactored after pair review


onboarding-task/src/components/AddItemElement.js
```diff
@@ -0,0 +1,37 @@
+import React, { Component } from 'react';
+
+class AddItemElement extends Component {
+  static displayName = 'AddItemElement';
+  static propTypes = { onAdd: React.PropTypes.func.isRequired };
```
PetrSvirak: You use member import on `Component` why not use it on `PropTypes` as well?

onboarding-task/src/components/ListItem.js
```diff
@@ -0,0 +1,69 @@
...
+  _labelClick() {
+    this.setState({ isEditable: true });
+  }
+
+  _inputChange(event) {
+    this.setState({ tempValue: event.target.value });
+  }
+
+  _cancelEdit() {
+    this.setState(() => ({ tempValue: this.props.item.value, isEditable: false }));
```
PetrSvirak: Why the function?

onboarding-task/src/components/ListItemForm.js
```diff
@@ -0,0 +1,29 @@
+import React, { Component } from 'react';
+
+class ListItemForm extends Component {
+  static displayName = 'ListItemForm';
```
PetrSvirak: It is actually not a form, is it? Even if there actually was a form, it is still not a good name for a component. Concentrate more on what component does, when would one use it, what is its purpose…

onboarding-task/src/components/ListItemForm.js
```diff
@@ -0,0 +1,29 @@
...
+    onCancel: React.PropTypes.func.isRequired,
+    onChange: React.PropTypes.func.isRequired };
+
+  render() {
+    const value = this.props.value;
+    return (
+      <div className="form-inline">
+        <div className="form-group">
+          {this.props.index}.
+          <input className="form-control" type="text" value={value} onChange={this.props.onChange} />
```
PetrSvirak: It is a pretty simple component and it would make sense to make it a [stateless functional component](https://hackernoon.com/react-stateless-functional-components-nine-wins-you-might-have-overlooked-997b0d933dbc#.vsqhcxxbs). However, should this component really be stateless? It seems to me that it is stateless only because its state (the value a user writes into the input box) gets stored by a component above it in the hierarchy. This hierarchically superior component does have quite no use for this state value, though. It would make more sense to me to either introduce state in this component or push the input's value all the way up (making it part of each item in the `List` component).

onboarding-task/src/components/ListItem.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  _saveValue() {
+    this.props.onEdit({ id: this.props.item.id, value: this.state.tempValue });
```
Suzii: If you want to only send `id` and `value`, I strongly suggest renaming the callback function to something that tells it is only responsible for changing the `value` of the item, e.g: `onItemValueEdit()`. Otherwise, change the implementation to `this.props.onEdit({ ...this.props.item, value: this.state.tempValue })` so that the other properties are not lost. I think the current implementation is nicely ready for other properties being added to the item without having props explosion to deal with.

onboarding-task/src/components/List.jsx
```diff
@@ -1,34 +1,58 @@
...
+    this.state = { itemsMap: {}, itemsToDisplay: [] };
+
+    this._addItem = this._addItem.bind(this);
+    this._editItem = this._editItem.bind(this);
+    this._deleteItem = this._deleteItem.bind(this);
+  }
+
+  _addItem(value) {
+    const newItem = { id: generateGUID(), value };
+    const itemsMap = Object.assign({}, this.state.itemsMap, { [newItem.id]: newItem });
```
davidcechak: Yes, Zuzka told me. I chose Object.assign() for educational purpose (to remember that it exist).
Well, it is even more educational to rewrite it once more. :smile:

PetrSvirak:
Nice work, thanks.

PR #19, opened 2017-02-23
Features/task 3

pavelf07:
Implemented task 3.


onboarding-task/src/components/List.jsx
```diff
@@ -1,75 +1,39 @@
...
+import React, { Component, PropTypes } from 'react';
+import ImmutablePropTypes from 'react-immutable-proptypes';
 
 import { ListItem } from './ListItem.jsx';
 import { AddForm } from './AddForm';
 import { EditForm } from './EditForm';
-import { generateId } from './../utils/idGenerator';
-import { Item } from '../models/ItemModel.js';
 
 class List extends Component {
```
PetrSvirak: One should extend [`PureComponent`](https://facebook.github.io/react/docs/react-api.html#react.purecomponent) to leverage immutables' advantage.

onboarding-task/src/containers/ListContainer.js
```diff
@@ -0,0 +1,38 @@
...
+    },
+  };
+};
+
+const ListContainer = connect(
+  mapStateToProps,
+  mapDispatchToProps,
+)(List);
+
+export { ListContainer };
```
PetrSvirak: I do not see a point in having higher order component suffixed with word `container` (while being imported from a file residing in the `containers` folder). Simply rename the file and exported function (removing the suffix).

onboarding-task/src/reducers/areEditableReducer.js
```diff
@@ -0,0 +1,17 @@
+import { Set } from 'immutable';
+import { START_EDIT_ITEM, STOP_EDIT_ITEM } from '../actions/actionTypes.js';
+
+const areEditable = (state = Set(), action) => {
```
PetrSvirak: Again, the name does not seem proper for a collection of whatever items (including, but not restricted to `id`s). I like the separation of concerns in the state, though – seems very clear to me.

onboarding-task/src/containers/ListContainer.js
```diff
@@ -0,0 +1,38 @@
+import { connect } from 'react-redux';
+import { addItem, deleteItem, updateItem } from '../actions/listActionCreators.js';
+import { startEditItem, stopEditItem } from '../actions/areEditableActionCreators.js';
+import { List } from '../components/List.jsx';
+
+const mapStateToProps = (state) => {
+  return {
+    list: state.list,
```
PetrSvirak: Consider using items transformation in combination with memoization. Check [reselect](https://www.npmjs.com/package/reselect) out. In total, it might be a bit complex for the project of this size, yet considered where you are heading with your code, I thought it might be an interesting exercise for you.

onboarding-task/src/reducers/listReducer.js
```diff
@@ -0,0 +1,23 @@
+import { OrderedMap } from 'immutable';
+import { ADD_ITEM, DELETE_ITEM, UPDATE_ITEM } from '../actions/actionTypes.js';
+import { Item } from '../models/ItemModel.js';
+
+const list = (state = OrderedMap(), action) => {
+  switch (action.type) {
+    case ADD_ITEM:
+      return state.set(action.payload.id, Item({ id: action.payload.id, text: action.payload.text }));
```
PetrSvirak: It is not reducer's responsibility to create an item. This should be done by the action factory.

onboarding-task/test/areEditableReducer.test.js
```diff
@@ -0,0 +1,23 @@
+import { Set } from 'immutable';
+import { startEditItem, stopEditItem } from '../src/actions/areEditableActionCreators.js';
+import { areEditable } from '../src/reducers/areEditableReducer.js';
+
+describe('areEditableReducer', () => {
+  const testId = '0aeeaa2b-1a2a-482c-b2a6-b172109071e7';
+
+  it('should return initial state for undefined state', () => {
+    expect(areEditable(undefined, {})).toEqual(Set());
```
PetrSvirak: The object (action) you pass to a reducer has to [contain at least `type` property](http://redux.js.org/docs/basics/Actions.html).

onboarding-task/test/areEditableActionCreators.test.js
```diff
@@ -0,0 +1,25 @@
+import { startEditItem, stopEditItem } from '../src/actions/areEditableActionCreators.js';
+import { START_EDIT_ITEM, STOP_EDIT_ITEM } from '../src/actions/actionTypes.js';
+
+describe('editModeActionCreators', () => {
+  const testId = '0aeeaa2b-1a2a-482c-b2a6-b172109071e7';
+
+  it('returns start edit item action', () => {
+    expect(startEditItem(testId)).toEqual({
```
PetrSvirak: * It is a "one-liner", true, however, would it be much more readable if parts of `expect(…).toEqual(…);` where actually intention-named local variables and, as a result, the before-mentioned statement was actually on a single line?
* Aren't you actually testing that action creators work correctly in the reducers tests?

onboarding-task/test/areEditableReducer.test.js
```diff
@@ -0,0 +1,23 @@
...
+  });
+
+  it('should add item to areEditable set on START_EDIT_ITEM', () => {
```
PetrSvirak: I personally don't like this `mock` suffix. It might be just that I'm not accustomed to this name, however, in this case, I believe it is not a mock in the true sense of the word anyway: 

What you describe here is actually a (reducer's) state expected to be present after an action is sent to a reducer.  Prefixing such variable/constant with `expected` just seems a bit more precise. 
Storing a set retrieved from the reducer in a variable (or constant more precisely) suffixed with `After` or prefixed with `actual` seems just a bit more right. 
The initial state (the `Set()` present e.g. on line 15) could be stored in a constant prefixed with `initial` or suffixed with `Before` (and should definitely not be a call hidden in a call of a call of a function ;-)).

onboarding-task/test/areEditableReducer.test.js
```diff
@@ -0,0 +1,23 @@
...
+    let areEditableMock = Set();
+    areEditableMock = areEditableMock.add(testId);
```
PetrSvirak: When initializing a `Set`, items can be passed to it (avoiding at least one unnecessary immutable object creation).

onboarding-task/test/idGenerator.test.js
```diff
@@ -0,0 +1,8 @@
+import { generateId } from '../src/utils/idGenerator.js';
+
+describe('id Generator', () => {
```
PetrSvirak: 👍

onboarding-task/test/listReducer.test.js
```diff
@@ -0,0 +1,48 @@
...
+  const firstTestId = '0aeeaa2b-1a2a-482c-b2a6-b172109071e7';
+  const addItem = addItemFunctionCreator(() => firstTestId);
+
+  it('should return initial state', () => {
+    expect(list(undefined, {})).toEqual(OrderedMap());
+  });
+
+  it('should handle add item', () => {
+    let mockList = OrderedMap();
+    let testList = OrderedMap();
```
PetrSvirak: Just to prove my point a bit further. Forget the following implementation and that you wrote the tests. What would you say is the difference between `mockList` and `testList`? 😈

onboarding-task/src/viewModels/ViewItem.js
```diff
@@ -0,0 +1,9 @@
+import { Record } from 'immutable';
+
+const ViewItem = new Record({
+  id: undefined,
+  text: '',
+  isEdited: undefined,
```
PetrSvirak: * It is a boolean property, `false` should be the natural option here.
* There were also some concerns about naming this property, however, I believe that, in given circumstances, it is OK to leave it as it is (renaming the property would inevitably result in renaming action creators and reducer).

onboarding-task/src/components/List.jsx
```diff
@@ -78,13 +41,13 @@ class List extends Component {
         <div className="col-sm-12 col-md-offset-2 col-md-8">
           <pre>
```
PetrSvirak: I have overlooked this one before; removing the tag made for code display will make your UI nicer ;-)

PR #23, opened 2017-03-03
Features/task 4

pavelf07:
App is rewritten to typescript.


onboarding-task/src/reducers/itemOrder.ts
```diff
@@ -0,0 +1,18 @@
+import { OrderedSet } from 'immutable';
+import { ADD_ITEM, DELETE_ITEM } from '../actions/actionTypes';
+import { IAction } from '../actions/IAction';
+
+const itemOrder = (state = OrderedSet(), action: IAction) => {
```
PetrSvirak: It should be plural, shouldn't it?

onboarding-task/src/selectors/getViewItems.ts
```diff
@@ -1,13 +0,0 @@
-const memoize = require('memoizee');
```
PetrSvirak: You lost file history. Be more careful on renaming, it can sometimes take an extra commit, however, with longer and longer history, it worths it.

onboarding-task/src/containers/Item.ts
```diff
@@ -0,0 +1,36 @@
+const { connect } = require('react-redux');
```
Suzii: And do not use require at all, have a look at the updated code conventions on how to import defaults in typescript :)

onboarding-task/src/containers/List.ts
```diff
@@ -1,24 +1,22 @@
...
 
-const mapDispatchToProps = (dispatch: Dispatch<IAction>) => {
+const mapDispatchToProps = (dispatch: Dispatch) => {
   return {
     addItem: (text: string) => dispatch(addItem(text)),
   };
 };
 
-const ListContainer: React.ComponentClass<undefined> = connect(
+const ListContainer = connect(
```
Suzii: Why did you delete `React.ComponentClass<undefined>`? Like this, any props could be sent to container and will be ignored.

PR #24, opened 2017-03-06
Allow to test .ts files

pavelf07:
Also works writing tests in typescript, but Typescript throws error on
jest (describe, it, etc.)


Suzii:
Awesome stuff, thanks a lot!! :)

PR #25, opened 2017-03-07
Features/task 4

ivanjrn:


onboarding-task/src/interfaces/IItemAction.ts
```diff
@@ -0,0 +1,6 @@
+import { IPayload } from './IPayload';
+
+export interface IItemAction {
+    type: string;
+    payload?: IPayload | any;
```
PetrSvirak: Is there any action of yours that actually does not have a payload? (I admit I actually might have missed some) If not, make the `payload` required. Also, I understand why you used `IPayload`, however, you would have to specify all action's interfaces and make this one generic or at least name all interfaces here. This is quite an overkill and plenty of work to do (assuming you test actions together with reducers). Just leave the `payload` to be of `any` type. Also, the naming here is a bit misleading as the `IPayload` does not suggest in any way a consumer should expect `id` and `text` property in it, moreover both optional.

onboarding-task/src/reducers/item.ts
```diff
@@ -1,16 +1,20 @@
...
   switch (action.type) {
-    case ADD_ITEM: return new ItemRecord({
+    case ADD_ITEM:
+      return new ItemRecord({
         guid: action.payload.guid,
         text: action.payload.text,
       });
-    case UPDATE_ITEM_TEXT: return state.set('text', action.payload.text);
-    default: return state;
+    case UPDATE_ITEM_TEXT:
```
PetrSvirak: Place empty line between one case end and new case beginning for better readability, please.

onboarding-task/src/reducers/itemsFlags.ts
```diff
@@ -1,19 +1,22 @@
...
-    case DELETE_ITEM: return state.delete(action.payload.guid);
+    case ADD_ITEM:
+      return state.set(action.payload.guid, { isEdited: false });
+    case DELETE_ITEM:
+      return state.delete(action.payload.guid);
     case UPDATE_ITEM_TEXT:
-    case TOGGLE_EDIT_MODE: return state.set(action.payload.guid, {'isEdited': !state.get(action.payload.guid).isEdited});
-    default: return state;
+    case TOGGLE_EDIT_MODE:
+      return state.set(action.payload.guid, { isEdited: !state.get(action.payload.guid).isEdited });
```
PetrSvirak: Extract original state into a separate variable for better readability, please.

onboarding-task/src/reducers/itemsById.ts
```diff
@@ -1,17 +1,21 @@
...
-import * as Immutable from 'immutable';
 import { item } from './item';
-import {IItemAction} from '../actions/actionCreators';
+import { IItemAction } from '../interfaces/IItemAction';
+import { Map } from 'immutable';
+import { IItemRecord } from '../interfaces/IItemRecord';
 
-const map = Immutable.Map();
-const itemsById = (state = map, action: IItemAction) => {
+const map: Map<string, IItemRecord> = Map<string, IItemRecord>();
```
PetrSvirak: I don't like the name `map` here - isn't it something more like an `emptyItemsById` (seems a bit more descriptive)? Also, I believe that you mustn't specify a type that can be derived from declaration - simple `= Map<string, IItemRecord>();` is enough to know what `map` constant is type of.

onboarding-task/src/reducers/itemsOrder.ts
```diff
@@ -0,0 +1,18 @@
...
+import { List } from 'immutable';
+
+const list: List<string> = List<string>();
+function itemsOrder (state: List<string> = list, action: IItemAction): List<string> {
+  switch (action.type) {
+    case ADD_ITEM:
+      return state.push(action.payload.guid);
+    case DELETE_ITEM:
+      const index: number = state.indexOf(action.payload.guid);
+      return state.splice(index, 1) as List<string>;
```
PetrSvirak: `Immutable.List` actually has method `remove` that accepts index as an argument. Also, I would personally prefer `OrderedSet` here because you do not actually work with order (indexes) anywhere in the code, you only expect the order to be present and each item to exist only once (actually) and thus you might find a `set` more convenient.

onboarding-task/src/interfaces/IItemAction.ts
Suzii: I think it should be optional (may be undefined) and agree with the fact that IPayload interface is an overkill.

onboarding-task/src/reducers/itemsById.ts
Suzii: ⬆️  👍

PetrSvirak:
Just to clarify (after discussion with @Suzii): what I meant referring to the `.idea` folder was that your task has, in a sense, nothing to do with the changes in the folder you made.

_WebStorm_ has, quite unfortunately from MPOV, mixed project settings and user settings under a single folder, thus it might be tricky to distinguish what parts of it are supposed to be in your repository and what part should not be there. This only adds to the reasons behind my opinion that an explicit task, or at least explicit commit, should be dedicated to amendments like these.

I believe (and do correct me if I am wrong) that your very changes in the folder were driven by `git add *` rather the fully conscious decision and that triggered my initial notice because you always have to be aware of and understand to every single file and every single change you commit.

onboarding-task/src/components/AddItem.tsx
```diff
@@ -23,14 +23,12 @@ class AddItem extends React.PureComponent<IAddItemProps, IAddItemState> {
     this._createNewItem = this._createNewItem.bind(this);
   }
 
-  _onInputChange(e: any) {
+  _onInputChange(e: React.ChangeEvent<HTMLInputElement>) {
```
Suzii: So as suggested you should use `React.SyntheticEvent<HTMLInputElement>`

onboarding-task/src/components/List.tsx
```diff
@@ -28,7 +35,7 @@ class List extends Component {
         <div className="col-sm-12 col-md-offset-2 col-md-8">
           <table className="table table-bordered">
             <tbody>
-              {items.map((guid, index) =>
+              {items.map((guid: string, index: number) =>
```
Suzii: Just FYI, I don't think explicite types are necessary here.

onboarding-task/src/utils/itemRecord.ts
```diff
@@ -0,0 +1,29 @@
...
+  readonly guid: string;
+  readonly text: string;
+}
+
+class ItemRecord extends Immutable.Record(defaultItem) implements IItem {
+  readonly guid: string;
+  readonly text: string;
+}
+
+const ViewItemRecord = Immutable.Record({
```
Suzii: You most likely forgot to make this TypeScript-ish. Anyway there is no need for ViewModels to use immutable.js. Just make it an interface. Also please move this all into `models/` folder and create separate files for `ItemRecord` and `IItemViewModel`.

Suzii:
Nicely done, thanks a lot! :)

PR #26, opened 2017-03-07
Features/task 4

JiriLojda:


onboarding-task/src/components/ListItemLabel.tsx
```diff
@@ -0,0 +1,10 @@
+import React = require('react');
+
+import { IAction } from '../interfaces/IAction';
+
+type onClickType = () => IAction;
+
+const ListItemLabel = ({ text, index, onClick }: { text: string; index: number; onClick: onClickType }) =>
```
Suzii: Create props interface to make it a bit more readable.

onboarding-task/src/actionCreators/createListItemFactory.ts
```diff
@@ -0,0 +1,24 @@
+import { CREATE_ITEM_IN_LIST } from '../constants/actionTypes';
+
+interface ICreateListItemAction {
```
Suzii: Anyway, see the review summary - i think you can use IAction basically everywhere.

PR #27, opened 2017-03-14
Feature/task 1

sutrkiller:
Ready for review

### DIFF
--- a/onboarding-task/src/components/CreateItem.jsx
+++ b/onboarding-task/src/components/CreateItem.jsx
@@ -1,6 +1,11 @@
-import React from 'react';
+import React, { PureComponent, PropTypes } from 'react';
 
-class CreateItem extends React.Component {
+class CreateItem extends PureComponent {
+  static displayName = 'CreateItem';
+  static propTypes = {
+    onItemAdd: PropTypes.func.isRequired,
+  };
+
   constructor(props) {
     super(props);
 
@@ -8,36 +13,39 @@ class CreateItem extends React.Component {
       text: '',
       error: '',
     };
-    this._handleSubmit = this._handleSubmit.bind(this);
-    this._handleEditText = this._handleEditText.bind(this);
   }
 
-  _handleEditText(event) {
-    this.setState({ text: event.target.value });
-  }
+  _handleEditText = (event) => {
+    this.setState({
+      text: event.target.value,
+    });
+  };
 
-  _handleSubmit(event) {
-    if (this.props.onItemAdd(this.state.text)) {
-      this.setState({ text: '' });
-    }
-    else {
-      this.setState({ error: 'Enter non-empty value!' });
-    }
+  _handleSubmit = (event) => {
     event.preventDefault();
-  }
+    const text = this.state.text;
+    if (!text.match(/\S/)) {
+      this.setState({
+        error: 'Enter non-empty value!',
+      });
+      return;
+    }
+    this.props.onItemAdd(text);
+    this.setState({
+      text: '',
+    });
+  };
 
   render() {
     return (
       <form onSubmit={this._handleSubmit} className="form-inline">
         <input type="text" className="form-control" value={this.state.text} onChange={this._handleEditText} />
-        <input type="submit" className="btn btn-default" value="Add" />
+        <button type="submit" className="btn btn-default"> Add</button>
         <br />
-        <span style={{ color: 'red' }}>{this.state.error}</span>
+        <span className="text-danger">{this.state.error}</span>
       </form>
     );
   }
 }
 
-CreateItem.propTypes = { onItemAdd: React.PropTypes.func };
-
-export default CreateItem;
+export { CreateItem };
--- a/onboarding-task/src/components/CreateItem.jsx
+++ b/onboarding-task/src/components/CreateItem.jsx
@@ -20,9 +20,10 @@ class CreateItem extends PureComponent {
     };
   }
 
-  _onTextChange = (event) => {
+  _onTextChange = (value, isValid) => {
     this.setState({
-      text: event.target.value,
+      text: value,
+      isAddDisabled: !isValid,
     });
   };
 
@@ -36,16 +37,10 @@ class CreateItem extends PureComponent {
     });
   };
 
-  _onValidityChange = (isValid) => {
-    this.setState({
-      isAddDisabled: !isValid,
-    });
-  };
-
   render() {
     return (
       <form onSubmit={this._handleSubmit} className="form-inline">
-        <Input value={this.state.text} onChange={this._onTextChange} validate={validateItemText} onValidityChange={this._onValidityChange} />
+        <Input value={this.state.text} onChange={this._onTextChange} validate={validateItemText} />
         <button type="submit" className="btn btn-default" disabled={this.state.isAddDisabled}>Add</button>
       </form>
     );
--- a/onboarding-task/src/components/CreateItem.jsx
+++ b/onboarding-task/src/components/CreateItem.jsx
@@ -8,16 +8,16 @@ class CreateItem extends React.Component {
       text: '',
       error: '',
     };
-    this.handleSubmit = this.handleSubmit.bind(this);
-    this.handleChange = this.handleChange.bind(this);
+    this._handleSubmit = this._handleSubmit.bind(this);
+    this._handleEditText = this._handleEditText.bind(this);
   }
 
-  handleChange(event) {
+  _handleEditText(event) {
     this.setState({ text: event.target.value });
   }
 
-  handleSubmit(event) {
-    if (this.props.onItemAdded(this.state.text)) {
+  _handleSubmit(event) {
+    if (this.props.onItemAdd(this.state.text)) {
       this.setState({ text: '' });
     }
     else {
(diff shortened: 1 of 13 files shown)


onboarding-task/src/components/CreateItem.jsx
```diff
@@ -0,0 +1,51 @@
...
+    });
+  };
+
+  render() {
+    return (
+      <form onSubmit={this._handleSubmit} className="form-inline">
+        <input type="text" className="form-control" value={this.state.text} onChange={this._handleEditText} />
+        <button type="submit" className="btn btn-default"> Add</button>
+        <br />
+        <span className="text-danger">{this.state.error}</span>
```
Suzii: I would prefer if you conditionally rendered whole span element as probable should not be in the DOM at all if there is no error

onboarding-task/src/components/List.jsx
```diff
@@ -1,37 +1,89 @@
...
+  _onItemAdd = (text) => {
+    const id = _generateUUID();
+    const items = this.state.items;
+    const item = {
+      id,
+      text,
+      editing: false,
+    };
+    items.set(id, item);
+    this.setState({ items });
```
Suzii: If you have a longer method (like this one), put an empty line before calling `setState()`. It basicaly serves as a return in normal method.

onboarding-task/src/components/List.jsx
```diff
@@ -1,37 +1,89 @@
...
+  };
+
+  _updateItemFromState =(id, values) => {
+    const items = this.state.items;
+    const item = items.get(id) || {};
+    const newItem = {
+      ...item,
+      ...values,
+    };
+    items.set(id, newItem);
```
Suzii: You are trying to be immutable in this method (at least with `item`) but you directly mutate the `items` map. It is not a problem right now, react handles it correctly here, but it is a good practice not to do it :) And you should be deffinitely aware of it as it might be costy in a larger app where you use flux architecture. But as I said in this particular case it does not matte, just make sure to avoid it in next tasks.

onboarding-task/src/components/List.jsx
```diff
@@ -1,37 +1,89 @@
...
+    this.setState({ items });
+  };
+
   render() {
+    const listItems = [...this.state.items.values()].map((item, i) =>
```
Suzii: My eyes are bleeding when I see this :D I know, I know this is the least complicated way to do the job when you have Map...

onboarding-task/src/components/List.jsx
```diff
@@ -1,37 +1,89 @@
...
+      <ListRow key={item.id} index={i + 1} item={item} onItemClick={this._onItemClick} onItemUpdate={this._onItemUpdate} onItemDelete={this._onItemDelete} onItemCancel={this._onItemCancel} />
```
Suzii: Put each prop on a new line to make the code more readable.

onboarding-task/src/components/ListRowEdit.jsx
```diff
@@ -0,0 +1,62 @@
...
+  };
+
+  _onItemCancel = (event) => {
+    event.preventDefault();
+    this.props.onItemCancel(this.props.item.id);
+  };
+
+  render() {
+    return (
+      <div className="list-group-item">
```
Suzii: I would move this tag to `List.jsx` component. You can decide later to change the way items are rendered to list - for example you will want to use `<ul>` or event `<table>`. When you have all the logic of "rendering list of something" in one component it is easier to maintain and make changes.

onboarding-task/src/components/ListRow.jsx
```diff
@@ -0,0 +1,28 @@
...
+    index: PropTypes.number.isRequired,
+    item: PropTypes.object.isRequired,
+    onItemClick: PropTypes.func.isRequired,
+    onItemUpdate: PropTypes.func.isRequired,
+    onItemDelete: PropTypes.func.isRequired,
+    onItemCancel: PropTypes.func.isRequired,
+  };
+
+  render() {
+    let row = null;
```
Suzii: Why not returning immediately?

onboarding-task/src/components/ListRow.jsx
```diff
@@ -0,0 +1,28 @@
...
+    if (this.props.item.editing) {
+      row = <ListRowEdit index={this.props.index} item={this.props.item} onItemUpdate={this.props.onItemUpdate} onItemDelete={this.props.onItemDelete} onItemCancel={this.props.onItemCancel} />;
```
Suzii: Put each prop on new line. Same for line 22.

onboarding-task/src/components/ListRowDisplay.jsx
```diff
@@ -0,0 +1,28 @@
...
+
+  _onItemClick = (event) => {
+    event.preventDefault();
+    this.props.onItemClick(this.props.item.id);
+  };
+
+  render() {
+    return (
+      <a href="#" className="list-group-item" onClick={this._onItemClick}>
+        {this.props.index}. {this.props.children}<span>{this.props.item.text}</span>
```
Suzii: `this.props.children` why?

onboarding-task/src/components/ListRowEdit.jsx
```diff
@@ -0,0 +1,62 @@
...
+
+  _onTextChange = (event) => {
+    this.setState({
+      text: event.target.value,
+    });
+  };
+
+  _onItemUpdate = (event) => {
+    event.preventDefault();
+    if (!this.state.text.match(/\S/)) {
```
Suzii: You have this logic twice, the other usage is in `CreateItem.jsx`. Wouldn't it be worth it to extract this business logic to one place? Some itemValidation file?

onboarding-task/src/components/CreateItem.jsx
```diff
@@ -41,8 +49,13 @@ class CreateItem extends PureComponent {
       <form onSubmit={this._handleSubmit} className="form-inline">
         <input type="text" className="form-control" value={this.state.text} onChange={this._handleEditText} />
         <button type="submit" className="btn btn-default"> Add</button>
-        <br />
-        <span className="text-danger">{this.state.error}</span>
+        {this.state.error &&
+          <div>
+            <br />
```
Suzii: Is `<br />` tag necessary?

onboarding-task/src/components/ListRowEdit.jsx
```diff
@@ -28,7 +29,7 @@ class ListRowEdit extends PureComponent {
 
   _onItemUpdate = (event) => {
     event.preventDefault();
-    if (!this.state.text.match(/\S/)) {
+    if (validateItemText(this.state.text).errors) {
       return;
```
Suzii: What happens if it is not valid? Shouldn't user at least see some message? Or disable button at all?

onboarding-task/src/utils/itemValidator.js
```diff
@@ -0,0 +1,11 @@
+function validateItemText(text) {  // might be part of validateItem() later, if other params are added
+  if (!text.match(/\S/)) {
+    return {
+      errors: true,
+      message: 'Enter non-empty text',
```
Suzii: I think `messages` array would be better here, since you might want to display context specific messages like 'item text should be more then 3 chars...'. And you wan to get all the messages and display for example only first? I would also probable rename `errors` to `isValid`, but thats just me and I am already kind of influenced by how Draft handles validation.

onboarding-task/src/utils/itemValidator.js
```diff
@@ -0,0 +1,11 @@
+function validateItemText(text) {  // might be part of validateItem() later, if other params are added
```
Suzii: Najsss.

sutrkiller:
Sorry for another large commit. I forgot again.

onboarding-task/src/components/Input.jsx
```diff
@@ -0,0 +1,88 @@
...
+  _onChange = (event) => {
+    this.props.onChange(event);
+    const validation = this.props.validate(event.target.value);
+    const errors = validation.isValid ? [] : validation.messages;
+    this.props.onInvalid(validation.isValid);
+
+    this.setState({ errors });
+  };
+
+  _getStyles = (hasErrors, isEmpty) => {
```
Suzii: Have a look at [classnames](https://www.npmjs.com/package/classnames) package. It is widely used an it helps in exactly these situations to make the css classes conditionals simpler :)

onboarding-task/src/components/Input.jsx
```diff
@@ -0,0 +1,88 @@
...
+
+  constructor(props) {
+    super(props);
+    this.state = { errors: [] };
+  }
+
+  _onChange = (event) => {
+    this.props.onChange(event);
+    const validation = this.props.validate(event.target.value);
+    const errors = validation.isValid ? [] : validation.messages;
```
Suzii: Why is the ternary operator needed here?

onboarding-task/src/components/Input.jsx
```diff
@@ -0,0 +1,88 @@
...
+    this.props.onInvalid(validation.isValid);
```
Suzii: I'm not sure why the callback is called `onInvalid` when it's called on every change, not only on invalidation.

Suzii:
Wonderful, thanks! :)

PR #32, opened 2017-03-20
Feature/task 1

LindaH32:
KA-109 - task 1 ready for review


onboarding-task/src/components/List.jsx
```diff
@@ -1,32 +1,129 @@
...
+import AddLine from './AddLine.jsx';
+import LineEdit from './LineEdit.jsx';
+import LineRead from './LineRead.jsx';
+import createGuid from '../GuidHelper.js';
 
 class List extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      lineTexts: [],
```
Suzii: Try to rename `lineTexts` to something which would indicate it contains objects not only strings like it does now.

onboarding-task/src/components/List.jsx
```diff
@@ -1,32 +1,129 @@
...
+    };
+  }
+
+  _handleAddLine = (text) => {
+    const editedLines = this.state.lineTexts
+      .concat([{
```
Suzii: Extract to a variable to make the code a bit more readable.

onboarding-task/src/components/List.jsx
```diff
@@ -1,32 +1,129 @@
...
+        text,
+        isEdited: false,
+      }]);
+
+    this.setState({ lineTexts: editedLines });
+  };
+
+  _handleDeleteLine = (lineId) => {
+    this.setState(
+      {
```
Suzii: Extract variable.

onboarding-task/src/components/List.jsx
```diff
@@ -1,32 +1,129 @@
...
+        lineTexts: this.state.lineTexts
+          .slice()
+          .filter((line) => line.id !== lineId),
+      }
+    );
+  };
+
+  _handleDoubleClick = (id) => {
```
Suzii: This whole method is a bit long and looks really messy due to formatting. Try to reformat the code a bit, put some new lines, extract updated item variable, etc. :)

onboarding-task/src/components/List.jsx
```diff
@@ -1,32 +1,129 @@
...
+    const updatedItems = this.state.lineTexts.slice();
+    updatedItems[indexOfClickedItem] = Object
+      .assign(
+        {}, clickedItem, { isEdited: true }
+      );
+
+    this.setState({ lineTexts: updatedItems });
+  };
+
+  _handleClickSave = (item) => {
```
Suzii: Same as the above. Try to make the code more readable.

onboarding-task/src/components/List.jsx
```diff
@@ -1,32 +1,129 @@
...
+    const updatedItem = Object
+      .assign(
+        {}, item, { isEdited: false }
+      );
+    const updatedItems = items.slice();
+    updatedItems[items.indexOf(item)] = updatedItem;
+
+    this.setState(
+      {
+        lineTexts: updatedItems,
```
Suzii: If it contains only one (max 2) properties, no need to put it on a new line. It is also inconsistent with previous methods where you have same thing on a single line.

onboarding-task/src/components/List.jsx
```diff
@@ -1,32 +1,129 @@
...
+      }
+    );
+  };
+
+  _renderLines = () => {
+    const items = this.state.lineTexts.map((line, index) => (
+      line.isEdited
+        ? <LineEdit
```
Suzii: The ternary operator should be generally used when there is simple logic (this is okay) but IMHO there should not be much code and it is also used inside of a lambda. Tat is too much stuff happening in one place. I would suggest you put the mapping logic directly to `render()` method and rename this method to `_renderLine()` where you would only return the line itself and get rid of the ternary operator.

onboarding-task/src/components/List.jsx
```diff
@@ -1,118 +1,83 @@
...
   _handleAddLine = (text) => {
-    const editedLines = this.state.lineTexts
-      .concat([{
-        id: createGuid(),
-        text,
-        isEdited: false,
-      }]);
-
-    this.setState({ lineTexts: editedLines });
+    const lines = this.state.rows;
```
Suzii: In state you renamed the property it would be nice to reflect the change throughout the whole app to keep naming consistent.

onboarding-task/src/components/List.jsx
```diff
@@ -1,118 +1,83 @@
...
+    const editedLines = lines.concat([{ id: createGuid(), text, isEdited: false }]);
```
Suzii: Extract new item to separate variable.

onboarding-task/src/components/List.jsx
```diff
@@ -1,118 +1,83 @@
...
-          text={line.text}
-          number={(index + 1)}
+  render() {
+    const rows = this.state.rows;
+    const renderedRows = rows.map((row, index) => (
+      <li className="list-group-item" >
+        <ListItem
+          key={row.id}
+          line={row}
+          number={index}
```
Suzii: Number? What does it mean? It's too general, how do I know as a user of the component what to pass there?

onboarding-task/src/components/List.jsx
```diff
@@ -1,118 +1,83 @@
...
-        {items}
-        <AddLine onAdd={this._handleAddLine} />
-      </ul>
-    );
-  };
-
-  render() {
+      </li>
+    ));
     return (
```
Suzii: Insert empty line before return.

PR #35, opened 2017-03-22
Feature/task 3

sutrkiller:
Ready for review


onboarding-task/src/reducers/itemsReducer.js
```diff
@@ -0,0 +1,32 @@
...
+      return prevState.set(action.value.id, newItem);
+    }
+    case ITEM_UPDATE: {
+      const updatedItem = itemReducer(prevState.get(action.value.id), action);
+      return prevState.set(action.value.id, updatedItem);
+    }
+    case ITEM_DELETE:
+      return prevState.delete(action.value.id);
+    case ITEM_TOGGLE_EDIT: {
+      // change locally using getIn/setIn??
```
PetrSvirak: It depends, since you have `item` reducer, it is OK for toggling to be in it. You might have considered this for `ITEM_CREATE` as it is actually more of collection operation (same as `ITEM DELETE`), but here you change the item itself. (not taking into account that the `editing` should not be part of `item` in the state)

onboarding-task/src/reducers/itemReducer.js
```diff
@@ -0,0 +1,24 @@
...
+  ITEM_CREATE,
+  ITEM_UPDATE,
+  ITEM_TOGGLE_EDIT,
+} from '../actions/actionTypes.js';
+import { Item } from '../models/Item.js';
+
+const itemReducer = (prevState = new Item(), action) => {
+  switch (action.type) {
+    case ITEM_CREATE:
+      return new Item({
```
PetrSvirak: You should not create an item in the reducer – it gives it more responsibilities and dependencies. It is the action that should actually create the item. Only very simple logic, if any, should be placed in the reducers as their main concern is the state transformation (and actions concern is to carry all the information necessary to all them to change the state)

onboarding-task/test/reducers/itemReducer.test.js
```diff
@@ -0,0 +1,63 @@
...
+    const expectedState = new Item({
+      id,
+      text: newText,
+    });
+    const resultState = itemReducer(new Item({ id, text }), updateItem(id, newText));
+
+    expect(resultState).toEqual(expectedState);
+  });
+
+  it('toggles editing correctly using toggleEditItem', () => {
```
PetrSvirak: You should actually have 2 tests here; for the case where `editing` is `true` and the other for the `false` (it is so simple for us developers to expect the things to work as we write them, isn't it)

onboarding-task/src/containers/ListContainer.js
```diff
@@ -0,0 +1,35 @@
...
+const mapStateToProps = (state) => {
+  return { items: state.items };
+};
+
+const mapDispatchToProps = (dispatch) => {
+  return {
+    onItemAdd: (text) => {
+      dispatch(createItem(text));
+    },
+    onItemClick: (id) => {
```
Suzii: I think that it's okay like this. If we did move it, we could easily end up with a container for each component in app.

onboarding-task/src/reducers/itemReducer.js
Suzii: I actually do not completely agree with this. Action tells that a new Item shall be created with text 'abc' and it should be reducers responsibility to create that item. Only reducer knows how the item shall be stored. If you decide to change the shape of the state, you would have to change both action creator and reducer. Like this, only your reducer is concenred about in which form data is actualy stored... E.g. if he splits the 'data' and 'ui properties', how does the action creator know what to actually dispatch? If he decides not to split the data and ui stuff (not recommended though) shall action creator dispatch item with ui data as well?

It's just my opinion and I'm open to discussion here.

onboarding-task/src/containers/ListContainer.js
Suzii: My bad sorry, I thought we were talking about moving the function from `ListRowContainer` to some not yet existing `ListRowDisplayContainer` (that would be an overkill). But here, Peter is right, it is absolutely useless in `ListContainer` and should be deffinitely moved to `ListRowContainer` along with deleting other unused props.

onboarding-task/src/reducers/itemReducer.js
PetrSvirak: What you say makes sense (well, as it alway does). My thoughts on this were based on too few previous reviews I guess. I have but one problem with your approach: should not `Id` then be also generated within the reducer? I mean what justifies the decision that `Id` is the responsibility of an action and `Item` is not? 
I understand that it would lead to the reducer effectively triggering another action on itself (what is completely senseless), yet then again should one have 2 semi-intelligent concepts in the communication chain (actions and reducers) or should one of them be dumb to ease the decision where each and every new logic should be put? I would personally somewhat prefer to have just a "single point of failure" and based on this logic, I would thus expect action to form both the object that is subject the data properties and the object that is subject to UI properties. I believe that it is actually one and only action creator who knows what needs to be done (it distributes responsibility and should make reducer pretty, but may be too much, thin). But it might just be me and too rigorous approach on the topic.
I also see at least one flaw in the what I wrote above: Reducers return the default state if none is provided to them. Pushing reducers pattern to its limit (theoretically, not practically – it would be useless), one ends up with a tree-like structure where each reducer either calls some other reducers (at least two) or cares exactly about one action ("simpelon leaf pseudo-reducer" you may call it). Now, one would have `create` reducer-ish function that would obtain literally no (previous) state and an action with `id` in its payload and would return a new object with default UI properties (similarly for data `item`). It would not then make sense to send anything but data required for the inner structure to be (re)created within the payload, just as you say.
So, you decide @Suzii :)

PetrSvirak:
Nice :+1:

PR #36, opened 2017-03-22
KA-110 - Feature/task 2

LindaH32:


onboarding-task/src/components/List.jsx
```diff
@@ -9,71 +10,63 @@ class List extends Component {
...
     };
   }
 
   _handleAddLine = (text) => {
     const lines = this.state.lines;
-    const newItem = { id: createGuid(), text, isEdited: false };
-    const editedLines = lines.concat([newItem]);
+    const id = createGuid();
+
+    const newItem = Map({ [id]: Map({ text, isEdited: false }) });
```
Suzii: This is not really correct. `newItem` is only this part: `Map({ text, isEdited: false })`. Fix and use `set()` on the next line instead of merge (it should be even more efficient operation).

onboarding-task/src/components/List.jsx
```diff
@@ -9,71 +10,63 @@ class List extends Component {
...
 
   _handleDoubleClick = (id) => {
     const rows = this.state.lines;
-    const clickedItem = rows.find((row) => row.id === id);
-    const indexOfClickedItem = rows.indexOf(clickedItem);
 
-    const updatedItem = Object.assign({}, clickedItem, { isEdited: true });
-    const updatedItems = rows.slice();
-    updatedItems[indexOfClickedItem] = updatedItem;
+    const updatedItems = rows.setIn([id, 'isEdited'], true);
```
Suzii: Najs :+1:

onboarding-task/src/components/List.jsx
```diff
@@ -9,71 +10,63 @@ class List extends Component {
...
-    const indexOfClickedItem = rows.indexOf(clickedItem);
+    const itemId = item.id;
 
-    const updatedItem = Object.assign({}, clickedItem, {
-      isEdited: false,
-      text: item.text,
-    });
-    const updatedItems = rows.slice();
-    updatedItems[indexOfClickedItem] = updatedItem;
+    const updatedItem = Map({ 'isEdited': false, 'text': [item.text] });
```
Suzii: Do not create completely new item. If item contained more properties, they would get lost with this approach. Try using `updateIn()` or `mergeIn()` methods instead of lines 47+48.

onboarding-task/src/components/List.jsx
Suzii: Also, please store `id` within item as well.

onboarding-task/src/components/List.jsx
```diff
@@ -9,71 +10,63 @@ class List extends Component {
...
-    const renderedRows = rows.map((row, index) => (
+    let index = 1;
+    const renderedRows = rows.map((row, id) => (
       <li className="list-group-item">
         <ListItem
-          key={row.id}
-          line={row}
-          index={index}
+          key={id}
+          line={{ id, row }}
```
Suzii: Store `id` within `row` and send only row as prop.

onboarding-task/src/components/LineEdit.jsx
```diff
@@ -1,6 +1,8 @@
-import React, { Component, PropTypes } from 'react';
+import React, { PureComponent, PropTypes } from 'react';
+
+class LineEdit extends PureComponent {
+  static displayName = 'LineEdit';
 
-class LineEdit extends Component {
   static propTypes = {
     line: PropTypes.object.isRequired,
```
Suzii: Define the structure of the object with immutable prop types.

Suzii: Same for `LineRead`

PR #39, opened 2017-03-29
Feature/task 4

sutrkiller:


onboarding-task/src/containers/ListRow.ts
```diff
@@ -0,0 +1,54 @@
...
+  index: number;
+}
+
+const getItemUi = (state: IAppState, props: IListRowOwnProps) => state.items.byUiProperties.get(props.id);
+const getItem = (state: IAppState, props: IListRowOwnProps) => state.items.byId.get(props.id);
+const getIndex = (_: IAppState, props: IListRowOwnProps) => props.index;
+
+const getItemViewModel = createSelector(
+  getItem, getItemUi, getIndex,
+  (item: IItem, itemUi: IItemUi, index: number) => ({
```
Suzii: Specify `IItemViewModel` return type?

onboarding-task/src/components/List.tsx
```diff
@@ -1,12 +1,22 @@
...
+const ImmutablePropTypes = require('react-immutable-proptypes');
+import { OrderedSet } from 'immutable';
+import TsComponent from './TsComponent';
+import { CreateItem } from './CreateItem';
+import { ListRow } from '../containers/ListRow';
+import { IAction } from '../interfaces/IAction';
 
-function List(props) {
-  const listItems = props.itemIds.valueSeq().map((id, i) => {
+const { PropTypes } = React;
```
Suzii: Why? Either use `import { PropTypes } from 'react';` in addition to default import, or just use React.PropTypes in the file (once you have typescript, the initial problem why we wanted to import PropTypes and Component directly is dismissed, you have fill IDE suport even with React.PropTypes)

PR #41, opened 2017-04-05
Feature/task 3

LindaH32:


onboarding-task/src/components/containerComponents/ListItemContainer.js
```diff
@@ -0,0 +1,29 @@
+import { connect } from 'react-redux';
+import { ListItem } from '../ListItem.jsx';
+import { editItem, deleteItem, saveChangesToItem, cancelChangesToItem } from '../../actionCreators/actionCreators.js';
+
+
+const mapStateToProps = (state, props) => {
+  return { line: props.line };
```
Suzii: You do not need to do this, own props are passed to connected component by redux itself.

onboarding-task/src/components/containerComponents/ListItemContainer.js
```diff
@@ -0,0 +1,29 @@
...
+};
+
+const mapDispatchToProps = (dispatch, props) => {
+  return {
+    index: props.index,
```
Suzii: MapDispatchToProps is ONLY for callbacks that use dispatch. You should never pass data via this function. Moreover, same as with `line` above, you do not even need to pass index explicitly, redux does it do you.

onboarding-task/src/components/containerComponents/ListItemContainer.js
```diff
@@ -0,0 +1,29 @@
...
+import { ListItem } from '../ListItem.jsx';
+import { editItem, deleteItem, saveChangesToItem, cancelChangesToItem } from '../../actionCreators/actionCreators.js';
+
+
+const mapStateToProps = (state, props) => {
+  return { line: props.line };
+};
+
+const mapDispatchToProps = (dispatch, props) => {
+  return {
```
Suzii: No need for empty lines between the properties.

onboarding-task/src/components/containerComponents/ListItemContainer.js
```diff
@@ -0,0 +1,29 @@
+import { connect } from 'react-redux';
+import { ListItem } from '../ListItem.jsx';
+import { editItem, deleteItem, saveChangesToItem, cancelChangesToItem } from '../../actionCreators/actionCreators.js';
+
+
+const mapStateToProps = (state, props) => {
+  return { line: props.line };
+};
+
+const mapDispatchToProps = (dispatch, props) => {
```
Suzii: Get rid of block scope and return keyword by wrapping the object you return from lambda function in parenthesis.

onboarding-task/src/components/containerComponents/ListItemContainer.js
```diff
@@ -0,0 +1,29 @@
...
+  return {
+    index: props.index,
+
+    onDelete: (id) => dispatch(deleteItem(id)),
```
Suzii: No need for a component to send the id when calling this callback. You have it here, you know what the id is, so pass it yourself, do not require the component to do it. Moreover, the component shall be simplified a lot once you do so.

onboarding-task/src/components/containerComponents/ListItemContainer.js
Suzii: Do so for all such cases where you have immediate return in lambdas.

onboarding-task/src/components/containerComponents/ListContainer.js
```diff
@@ -0,0 +1,18 @@
+import { connect } from 'react-redux';
+import { List } from '../List.jsx';
+import { addItem } from '../../actionCreators/actionCreators.js';
+
+const mapStateToProps = (state) => {
+  return { lines: state.lines };
```
Suzii: Pass only IDs of lines to the `List` component, and make the `LineContainer` get the line from the store based on `id` from `ownProps`. Also, make sure to memoize all of this (use [memoizee](https://www.npmjs.com/package/memoizee) library). 

If you leave it like this, or don't memoize, by clicking on any Item (making the form visible == switching the `isEdited` flag), the `lines` Map is recreated, given new reference, and thus the whloe `List` component is rerenderd -- completely unnecessary. Only particular `Line` component needs to be rerendered in such case. If the list is really long, the performance sufferes.

Suzii: I get that it is pretty complex stuff. If you don't understand any of this, ask personaly next week. It is pretty important that you understand how it works under the hood.

onboarding-task/src/components/List.jsx
```diff
@@ -1,101 +1,45 @@
...
-    ));
-
-    return (
+import { ListItemContainer } from './containerComponents/ListItemContainer';
+
+const List = ({
+    lines,
+    onAddLine,
+  }) => {
+  List.displayName = 'List';
```
Suzii: Move outside of the function where you have the prop types declared. Holds for all other functional components.

onboarding-task/src/containers/ListItemContainer.js
```diff
@@ -1,11 +1,20 @@
...
 import { ListItem } from '../components/ListItem.jsx';
 import { enableEditItem, deleteItem, saveChangesToItem, cancelChangesToItem } from '../actionCreators/actionCreators.js';
+import memoize from 'memoizee';
+
+const itemViewModel = (item, index) => ({
+  ...item.toObject(),
+  index,
+});
+
+const memoizedItem = memoize(itemViewModel);
```
Suzii: 👍 Awesome!

Suzii:
Wonderful! Thanks! :)

PR #46, opened 2017-04-19
Finished Task 5

sutrkiller:


onboarding-task/src/reducers/itemsReducers/itemsByIdReducer.ts
```diff
@@ -24,7 +24,7 @@ const itemsByIdReducer = (state = OrderedMap<string, Item>(), action: IAction):
     }
 
     case GET_ITEMS_SUCCESS: {
-      const receivedItems = action.payload.items.map((val: IItem) => [val.id, new Item(val)]);
+      const receivedItems = action.payload.items.map((val: IItemServerModel) => [val.id, new Item(val)]);
```
Suzii: You did not define an explicit constructor for Item, so the default one from Record will be use. That one effectively ignores properties from IItemServerModel, that were not defined on default item. This might introduce subtle bugs in the future. Next time, be careful and specify the constructor explicitely.

onboarding-task/src/reducers/rootReducer.ts
```diff
@@ -4,11 +4,11 @@ import { IAction } from '../actions/IAction';
 import { IAppState } from './IAppState';
 import { errorsReducer } from './errorsReducer';
 
-type IRootReducer = (state: IAppState, action: IAction) => IAppState;
+type rootReducer = (state: IAppState, action: IAction) => IAppState;
```
Suzii: Convention is to name custom types UpperCamleCase.

onboarding-task/src/utils/urlAdresses.ts
```diff
@@ -0,0 +1 @@
+export const ENDPOINT_URL = 'api/v1/items';
```
Suzii: What endpoint? This could have been separated into 2 constants, one for `api/v1` and second that would concatenate it with `items`

PR #49, opened 2017-04-25
Feature/task 3

VlastimilM:


onboarding-task/src/components/List.jsx
```diff
@@ -1,105 +1,21 @@
...
-import { Map as ImmutableMap, List as ImmutableList } from 'immutable';
-import { ItemForm } from './CreateItemForm';
-import { ListItem } from './ListItem';
-import { Item } from '../models/Item';
-import { generatePseudoUniqueID } from '../utils/keyGenerator';
+import { CreateItemForm } from '../containers/CreateItemFormContainer';
+import { ListItem } from '../containers/ListItemContainer';
 import { ListGroup, ListGroupItem } from 'react-bootstrap';
 
 class List extends PureComponent {
```
Suzii: You added container that passed props, but component has no `PropTypes` defined.

onboarding-task/src/components/List.jsx
```diff
@@ -1,105 +1,21 @@
...
-    const updatedItems = this.state.items.delete(idToDelete);
-    const updatedOrderedKeys = this.state.orderedIds.filter(x => x !== idToDelete);
-
-    this.setState({
-      items: updatedItems,
-      orderedIds: updatedOrderedKeys,
-    });
-  };
-
   _createListItems = () => {
```
Suzii: You you move logic from this method into `render()` (the result can be stil stored in constant as it is now) you can refactor whole component into functional stateless component, which should be always favoured if possible.

onboarding-task/src/containers/ListContainer.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * Created by VlastimilM on 9.4.2017.
+ */
+import { connect } from 'react-redux';
+import { List } from '../components/List';
+
+const mapStateToProps = (state) => {
+  return {
+    items: state.items,
```
Suzii: Do not pass all items data. `OrderedIds` should be enough for `List` to render underlying components. Let `ItemContainer` get data for specific item from store itself, based on `ownProps.id`.

onboarding-task/src/containers/ListItemContainer.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Created by VlastimilM on 9.4.2017.
+ */
+import { connect } from 'react-redux';
+import { saveItem, deleteItem, startEditingItem, stopEditingItem, updateItemText } from '../actions/actionCreators';
+import { ListItem } from '../components/ListItem';
+
+const mapDispatchToProps = (dispatch) => {
+  return {
+    onSave: (id, text) => dispatch(saveItem(id, text)),
```
Suzii: Why should the underlying component pass the `id`, when you already have it here in `ownProps` (second argument of `mapDispatchToProps` callback function)? Like that, most of the prop functions will be parameterless and the underlying components will be simplified greatly/. (No method required, so some would be even refactorable to stateless functional components.).

onboarding-task/src/containers/ListItemContainer.js
```diff
@@ -0,0 +1,20 @@
...
+    onDelete: (id) => dispatch(deleteItem(id)),
+    onUpdate: (id, text) => dispatch(updateItemText(id, text)),
+    onCancel: (id) => dispatch(stopEditingItem(id)),
+    onEdit: (id) => dispatch(startEditingItem(id)),
+  };
+};
+
+const listItemContainer = connect(undefined, mapDispatchToProps)(ListItem);
```
Suzii: Use `mapStateToProps` to get the item and create a ViewModel for item that would contain all data (including index, that you have in `ownProps`). Pass whole ViewModel object to underlying components. Memoize it to gain better performance (if you are unsure why/how, ask :))

onboarding-task/test/reducers/orderedIdsReducer.test.js
```diff
@@ -0,0 +1,28 @@
+import { List as ImmutableList } from 'immutable';
+
+import { getOrderedIds } from '../../src/reducers/items/orderedIdsReducer';
+import { deleteItem } from '../../src/actions/actionCreators';
+import { addItemFactory } from '../../src/actions/actionCreatorsFactory';
+import { unknownAction } from '../actions/helperActions';
+
+describe('orderedIds reducer', () => {
```
Suzii: No test for returning initial state.

onboarding-task/src/components/InsertedListItem.jsx
```diff
@@ -1,26 +1,16 @@
-import React, { PureComponent, PropTypes } from 'react';
+import React, { PropTypes } from 'react';
 
-class InsertedListItem extends PureComponent {
-  static displayName = 'InsertedListItem';
+export const InsertedListItem = (props) => {
+  return (
+    <div>
+      <span onClick={() => props.onEdit()}>{props.item.index + 1}. {props.item.text}</span>
```
Suzii: No need for creating new lambda function here, applies for all functions without arguments.

Suzii:
Thanks for the amendments, the code looks a lot cleaner now! Two of your commits lack the jira number in commit message, please, do not forget next time.

I have some further comments before merging the task:
- Why did you create a record type for viewModel? this is not actually necessary at all, plain object is sufficient, as the underlying component that receives it should not influence anything if it accidentally mutates it. But does not hurt anybody right now, so you can probably leave it as it is.
- However `propTypes` for viewModel??? Why? This is completely misunderstood usage of prop types... :(
- Please name your reducer same as the property in state they are responsible for (with an optional '*Reducer' suffix) -- this has already been mentioned in the previous review :(
- Adding and editing items on Enter reloads the page, this does not seem as very user-friendly behaviour. Read about `preventDefault()` on forms/submit buttons.
- Reducer tests: `itemsByIdReducer` definitely should not use other reducers for computing expected state... only basic actions are allowed to be reused.
- IMHO, the `appReducer.tes.js` should be deleted.
- Delete all PropTypes imports from react as it has become deprecated with the new version of react. The prop types have been moved to a separate npm package - 'prop-types'. Use that one.

onboarding-task/test/reducers/orderedIdsReducer.test.js
Suzii: Applies for all other reducers.

onboarding-task/README.md
```diff
@@ -295,7 +295,7 @@ For example, this:
 .App {
   display: flex;
   flex-direction: row;
-  align-items: center;
+  align-itemsByIds: center;
```
PetrSvirak: …which is quite better experience than calling `git diff` prior any commit. Either way, you have to be absolutely sure about literally any single character change you commit (understand it, agree with it) – if you see any undesirable change or anything you consider "magic" in your stage area, do not commit at all and ask.

PR #50, opened 2017-05-11
Feature/task 5

davidcechak:


onboarding-task/src/components/List.tsx
```diff
@@ -4,29 +4,39 @@ const ImmutablePropTypes = require('react-immutable-proptypes');
...
 }
 
 const List: React.StatelessComponent<IListDataProps & IListCallbacksProps> = (props) => {
   const itemsList = props.itemsOrder.toIndexedSeq().map((id: string) => (
-    <li className="list-group-item" key={id}>
-      <ListItem id={id}/>
-    </li>)
+    <div className="row" key={id}>
+      <div className="col-sm-12 col-sm-offset-2 col-sm-8">
+        <li className="list-group-item" key={id}>
```
PetrSvirak: The `key` should not be necessary here…

onboarding-task/src/components/List.tsx
```diff
@@ -4,29 +4,39 @@ const ImmutablePropTypes = require('react-immutable-proptypes');
...
+      <ErrorViewerContainer />
       <ul className="list-group">
         {itemsList}
-        <li className="list-group-item">
-          <AddItem onAdd={props.onAddItem}/>
-        </li>
+        <div className="row">
+          <span className="col-sm-12 col-sm-offset-2 col-sm-8">
+            <li className="list-group-item">
+              <AddItem onAdd={props.onAddItem}/>
```
PetrSvirak: You actually copy pasted quite some code here. First, is there a reason for one having `span` within the `row div` and the other `div`? Second, just `DRY` (this is not a reference to any :seedling: :wink:).
Consider creating a common component that wraps component(s) needed for individual rows. There can be either HOC or component wrapping used here (read through this [SO](http://stackoverflow.com/questions/36960675/difference-between-using-a-hoc-vs-component-wrapping), please). You might come to something like this:
````
const listItemWrapper = <TProps extends {id?: string}>(Component: React.ComponentClass<TProps>, props?: TProps) => (
  <div className="row" key={props && props.id}>
    <div className="col-sm-12 col-sm-offset-2 col-sm-8">
        <Component {...props} />
      </li>
    </div>
  </div>);
````

File: onboarding-task/src/containers/ListItemContainer.ts [hunk 3]
```diff
@@ -18,16 +18,17 @@ function constructViewModel(item: Item, flags: ItemFlags, index: number): IItemV
     id: item.id,
     value: item.value,
     isInEditMode: flags.editMode,
+    isSavedInDatabase: flags.isSavedInDatabase,
     index,
-  }
+  };
```
Suzii: Not entirely ture, lambdas must have a semi-colon after their definition, unlike funcions :P He would save himself merly a pair of curly braces.

onboarding-task/src/reducers/itemsDataReducer.ts
```diff
@@ -2,24 +2,39 @@ import { Map } from 'immutable';
...
+    case POSITIVELY_CREATE_ITEM_LOCALLY:
+      return state.set(action.payload.ueid, itemReducer(undefined, action));
 
     case DELETE_ITEM:
       return state.delete(action.payload.id);
 
+    case ITEMS_FETCHING_SUCCEED:
+      return Map<string, Item>(
+        action.payload.items
+          .map((item: any) => [item.id, new Item(item)])
```
Suzii: I would insist on fixing this one, were it my review :confused:

onboarding-task/src/reducers/itemsOrderReducer.ts
```diff
@@ -1,20 +1,34 @@
...
+    case ITEMS_FETCHING_SUCCEED:
+      return OrderedSet<string>(
+        action.payload.items
+          .map((item: Item) => item.id)
+      );
+
+    case ITEM_POST_SUCCEED:
+      return state
+        .delete(action.payload.item.ueid)
+        .add(action.payload.item.id);
```
Suzii: I was wondering if you will spot this one in the review :smiling_imp:

onboarding-task/src/reducers/itemsReducer.ts
```diff
@@ -1,13 +1,17 @@
...
+import { IAction } from '../actions/IAction';
 
-const itemsReducer = combineReducers({
-  items: itemsDataReducer,
-  itemsOrder: itemsOrderReducer,
-  itemsDisplayFlags: itemsFlagReducer,
-});
-
-export { itemsReducer };
+export function itemsReducer(prevState: IAppState = {} as IAppState, action: IAction): IAppState {
```
Suzii: I do agree with the `items` state subtree, it makes the state wonderfully extensible. The tests should not be afftected at all IMHO, only the containers. As for the `combineReducers`, I suggested to write it manually, since it is typed this way and we could not figure out how to do it with combine reducer version.

onboarding-task/src/stores/Dispatch.ts
```diff
@@ -1,3 +1,14 @@
...
+//   & ((dispatch: Dispatch) => Promise<IAction>);
+
+type dispatchAction = (action: IAction) => IAction;
+type dispatchThunk = (thunkAction: ((dispatch: dispatchAction) => Promise<IAction>)) => Promise<IAction>;
+export type Dispatch = dispatchThunk & dispatchAction;
+
+// export type Dispatch = (action: IAction | Dispatch) => IAction | Promise<IAction>;
+
+// export type Dispatch = (action: IAction | Dispatch1) => IAction;
+// type Dispatch1 = (action: IAction | Dispatch1) => Promise<IAction> ;
```
Suzii: :woman_facepalming:

onboarding-task/src/components/AddItem.tsx
```diff
@@ -18,7 +18,7 @@ class AddItem extends React.PureComponent<IAddItemProps, IAddItemState> {
     this.state = { inputValue: '' };
   }
 
-  _inputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+  _inputChange = (event: any) => {
```
Suzii: :woman_facepalming:

onboarding-task/src/components/List.tsx
Suzii: No need for HoC here, styling is usually just handled by simple functional components, whete the css classes are encapsulated in the component and it just renders all its passed children. HoC would even make the JSX look wierd IMHO. Have a look at [react-bootstrap](https://react-bootstrap.github.io/components.html#page-layout) for an inspiration.

onboarding-task/src/containers/ListItemContainer.ts
PetrSvirak: He would not save himself semi-colon itself, but the trouble with it (since he would have there all the time :stuck_out_tongue: )

onboarding-task/src/components/List.tsx
PetrSvirak: I did not both actually (HOC and wrapping component you described), if it weren't for TS and given circumstances, I would totally agree, however, HOC was simpler here since you need to give the `key` property the top-level `div` (sometimes) and pass `props` to inner component (only for `AddItem`) – all this combined resulted in (IMHO) less cluttered when used HOC. Definitely try both @davidcechak as well (and choose what you like more) :)

onboarding-task/src/containers/ListLoaderContainer.ts
```diff
@@ -0,0 +1,25 @@
+import * as React from 'react';
+import { connect } from 'react-redux';
+
+import { fetchItems } from '../actions/actionCreators';
+import { IAppState } from '../stores/IAppState';
+import { Dispatch } from '../stores/Dispatch';
+import { IListLoaderCallbacksProps, IListLoaderDataProps, loader } from '../components/loader';
+import { List } from './ListContainer';
+
+function mapStateToProps(state: IAppState): IListLoaderDataProps {
```
davidcechak: I would not lambda them to be consistent across the app, but thanks to this remark  I noticed I am inconsistent. :)

onboarding-task/src/actions/postItemFactory.ts
```diff
@@ -1,34 +1,43 @@
...
+                         generateId: () => string,
+                         receivePostItemError: (error: Error, itemUeid: string) => IAction,
+                         receiveItemCreated: (json: Item) => IAction,
+                         parseResponse: (errorMessage: string) => (response: Response) => Promise<any>) =>
+  (value: string) => {
+    return (dispatch: Dispatch): Promise<IAction> => {
+      const item = new Item({ ueid: generateId(), value });
+      dispatch(positivelyCreateItemLocally(item));
+
+      if (!navigator.onLine) {
```
PetrSvirak: What the heck is this?
Anyway, it surely is dependency as can clearly be seen from your tests. If you insist on using this here, make it a dependency. But it should not be necessary. Fetch will surely fail if there is no connection, so you can solve by catching a failed promise, should you not?

onboarding-task/src/reducers/IItemsState.ts
```diff
@@ -4,12 +4,12 @@ import { Item } from '../models/Item';
 import { ItemFlags } from '../models/ItemFlags';
 import { ErrorMessage } from '../models/ErrorMessage';
 
-interface IAppState {
+interface IItemsState {
   items: Immutable.Map<string, Item>;
-  itemsOrder: Immutable.OrderedSet<string>;
-  itemsFlags: Immutable.Map<string, ItemFlags>;
+  order: Immutable.OrderedSet<string>;
+  flags: Immutable.Map<string, ItemFlags>;
```
PetrSvirak: See, here you have `items`, `order`, `flags` – having a `list` instead of `items` here would also make sense, then you could use something called `itemsListReducer` to reduce this `list` and `IItemsState` would be a valid name. Since you named it the other way around, you have to rename `IItemsState` to `IItemsListState` (do the rename, please).

onboarding-task/test/actions/actionCreators.test.js
```diff
@@ -33,6 +34,25 @@ describe('actionCreators', () => {
...
+  // Set up for postItemFactory tests
+  const receivePostItemErrorAction = (error, itemUeid) => ({
+    type: ITEM_SAVE_FAILED,
+    payload: {
+      id: id,
+      itemUeid,
+      message: error.message,
+    },
+  });
+  // const receivePostItemErrorMock = (error, itemUeid) => receivePostItemErrorAction(error, itemUeid);
```
PetrSvirak: Commented code does not belong to any source control :wink:

onboarding-task/test/actions/actionCreators.test.js
```diff
@@ -81,43 +101,66 @@ describe('actionCreators', () => {
...
+
+    const postItem = postItemFactory(fetch, () => ueid, receivePostItemErrorMock, receiveItemCreatedMock, parseResponseMock);
+
+    return postItem(value)(dispatch).then(() => {
+      expect(dispatch.mock.calls[1][0].payload.item).toEqual(item);
+    });
+  });
+
+  // This test will be fixed soon
+  xit('postItem correctly calls receivePostItemError as a second call in dispatch, with network connection set as offline', () => {
```
PetrSvirak: I hope it will :wink:

onboarding-task/src/actions/postItemFactory.ts
```diff
@@ -1,34 +1,43 @@
...
-          })
-          .then(parseResponse(': Item was not correctly saved on the server'))
-          .then((json: Item) => dispatch(receiveItemCreated(json)))
-          .catch((error: Error) => dispatch(receivePostItemError(error, item.ueid)));
-      };
+const postItemFactory = (fetch: Fetch,
+                         generateId: () => string,
+                         receivePostItemError: (error: Error, itemUeid: string) => IAction,
+                         receiveItemCreated: (json: Item) => IAction,
+                         parseResponse: (errorMessage: string) => (response: Response) => Promise<any>) =>
```
PetrSvirak: Imagine how greater would it be if all the dependencies were passed in as a single object with a type defined separately (in this file though). Then consider doing it. I will not insist just because the sprint end is closing.

onboarding-task/src/@types/global.d.ts
```diff
@@ -2,4 +2,8 @@ import { IAction } from '../actions/IAction';
 
 type dispatchAction = (action: IAction) => IAction;
 type dispatchThunk = (thunkAction: ((dispatch: dispatchAction) => Promise<IAction>)) => Promise<IAction>;
-export type Dispatch = dispatchThunk & dispatchAction;
+
+declare global {
+  type Dispatch = dispatchThunk & dispatchAction;
```
PetrSvirak: :+1:

pavelf07:
Personally, I prefer to create an interface for dependency object which is separate from method definition. For me, it increases code readability.

onboarding-task/test/actions/actionCreators.test.js
```diff
@@ -128,7 +140,14 @@ describe('actionCreators', () => {
     const fetch = () => Promise.resolve(item);
     const dispatch = jest.fn(action => action);
 
-    const postItem = postItemFactory(fetch, generateId, receivePostItemErrorMock, receiveItemCreatedMock, identityFunc);
+    //const postItem = postItemFactory(fetch, generateId, receivePostItemErrorMock, receiveItemCreatedMock, identityFunc);
```
pavelf07: Delete comment.

onboarding-task/test/actions/actionCreators.test.js
```diff
@@ -174,9 +209,15 @@ describe('actionCreators', () => {
       then: () => Promise.resolve(item),
     });
     const dispatch = jest.fn(action => action);
-    const receiveItems = items => ({type: ITEMS_FETCHING_SUCCEED, payload: {items: items}});
-
-    const fetchItem = fetchItemsFactory(fetch, identityFunc, receiveItems, identityFunc);
+    const receiveItems = items => ({ type: ITEMS_FETCHING_SUCCEED, payload: { items: items } });
+
+    //const fetchItem = fetchItemsFactory(fetch, identityFunc, receiveItems, identityFunc);
```
pavelf07: Delete comment.

onboarding-task/src/actions/fetchItemsFactory.ts
```diff
@@ -2,17 +2,18 @@ import { IAction } from './IAction';
 import { requestItems } from './actionCreators';
 import { API_VERSION_1, ITEMS } from '../constants/urls';
 
-const fetchItemsFactory = (fetch: Fetch,
-                           parseResponse: (errorMessage: string) => (response: Response) => Promise<any>,
-                           receiveItems: (json: any) => IAction,
-                           receiveItemsFetchingError: (error: Error) => IAction) => {
+const fetchItemsFactory = (dependencies: {
```
pavelf07: Create a separate interface for dependency object.

onboarding-task/src/actions/postItemFactory.ts
```diff
@@ -5,17 +5,19 @@ import {
 } from './actionCreators';
 import { API_VERSION_1, ITEMS } from '../constants/urls';
 
-const postItemFactory = (fetch: Fetch,
-                         generateId: () => string,
-                         receivePostItemError: (error: Error, itemUeid: string) => IAction,
-                         receiveItemCreated: (json: Item) => IAction,
-                         parseResponse: (errorMessage: string) => (response: Response) => Promise<any>) =>
+const postItemFactory = (dependencies: {
```
pavelf07: Create a separate interface for dependency object.

onboarding-task/test/actions/actionCreators.test.js
```diff
@@ -115,7 +121,13 @@ describe('actionCreators', () => {
...
-    const postItem = postItemFactory(fetch, generateId, receivePostItemErrorMock, receiveItemCreatedMock, parseResponseMock);
+    const postItem = postItemFactory({
+      fetch,
+      generateId,
+      receivePostItemError: receivePostItemErrorMock,
+      receiveItemCreated: receiveItemCreatedMock,
+      parseResponse: parseResponseMock
+    });
 
     expect.assertions(2);
```
pavelf07: This is not necessary. Separate expects by a semicolon.

onboarding-task/test/actions/actionCreators.test.js
```diff
@@ -95,7 +95,13 @@ describe('actionCreators', () => {
...
-    const postItem = postItemFactory(fetch, generateId, receivePostItemErrorMock, receiveItemCreatedMock, parseResponseMock);
+    const postItem = postItemFactory({
+      fetch: fetch,
+      generateId: generateId,
+      receivePostItemError: receivePostItemErrorMock,
+      receiveItemCreated: receiveItemCreatedMock,
+      parseResponse: parseResponseMock
+    });
 
     expect.assertions(2);
```
pavelf07: This is not necessary. Separate expects by a semicolon.

onboarding-task/src/actions/actionCreators.ts
```diff
@@ -66,10 +67,26 @@ const deleteError = (id: string): IAction => ({
...
+const postItemWithDependencies = postItemFactory({
+  positivelyCreateItemLocally,
+  fetch: catchedFetch,
+  generateId: generateGuid,
+  receivePostItemError: receivePostItemErrorWithDependencies,
+  receiveItemCreated,
+  parseResponse,
+});
 
 export {
```
PetrSvirak: There are 4 unused exports here, please find them and remove them :)

onboarding-task/src/utils/parseResponse.ts
```diff
@@ -1,10 +1,9 @@
 export const parseResponse = (errorMessage: string) => (response: Response): Promise<any> => {
-  if (response.ok) {
-    return response.json();
-  }
   return response.json().catch(() =>
     Promise.reject(new Error(response.statusText + ': ' + errorMessage))
-  ).then((data: any) =>
-    Promise.reject(new Error(response.statusText + ': ' + data.modelState.value.toString()))
+  ).then((data: any) => {
```
PetrSvirak: It's still a bit hard for me to navigate mainly because of formatting. Consider removing and unnecessary `return` statement and inlining variables to a string and formatting the method something like this:
````
export const parseResponse = (errorMessage: string) => (response: Response): Promise<any> => response
    .json()
    .catch(
      () => Promise.reject(new Error(`${response.statusText}: ${errorMessage}`)))
    .then(
      (data: any) => response.ok
        ? data
        : Promise.reject(new Error(`${response.statusText}: ${data.modelState.value.toString()}`)));
````
It might be just a matter of taste, though.

BTW: Is the `toString()` method really necessary? The value your back-end provides is quite surely a string…

PR #51, opened 2017-05-11
KA-106 Rewrite async tests using return

davidcechak:
to get better feedback for failed test
rewrote based on Jest documentation [https://facebook.github.io/jest/docs/tutorial-async.html#content](https://facebook.github.io/jest/docs/tutorial-async.html#content)

### DIFF
--- a/onboarding-task/test/dummyTests.test.js
+++ b/onboarding-task/test/dummyTests.test.js
@@ -14,25 +14,17 @@ describe('Promise resolving', () => {
       resolve(42);
     });
 
-    expect.assertions(1);
-    return myPromise.then(result => {
-      expect(result).toBe(42);
-    });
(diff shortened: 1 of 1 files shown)


PetrSvirak:
@davidcechak, you  used `expect(…) || expect(…)` quite often (e.g. [here](https://github.com/Suzii/kentico-onboarding-js/pull/50/files#diff-e650c89423cbb3a9826c05a6e474ff73R182) in your `Task 5` tests. Could you explore if there is a better option? Either way, it might be worthy be used within the sample tests.
Also, there are see some tests looking like `… => { return … }` – it would be better to omit redundant scoping and write simpy `… => …` in my opinion.

Suzii:
Thanks for this one! :)

PR #60, opened 2017-07-11
Feature/task 3

milan009:


onboarding-task/src/reducers/rootReducer.js
```diff
@@ -0,0 +1,7 @@
+import { listReducer } from './listReducer';
+
+export const rootReducer = (state = {}, action) => {
+  return {
+    list: listReducer(state.list, action),
```
PetrSvirak: @Suzii  I'm afraid the root of this problem lays elsewhere: There is the component called `List` in the app whose state, I believe, is represented in this reducer. @milan009 It does not make much sense to me to rename the reducer and not rename the `List` component or vice versa.

Suzii: Feel free to rename the copoment if you'd like. But the idea of having one-to-one mapping between parts of redux store and components is not a good one. You can have many components rendering data from same part of store in different forms. However, I do admit the relationship in this case is pretty obvious.

PetrSvirak: I hope I did not suggest anything like such mapping being a general rule. My point was that we render something we call "list" and provide no backup story in the assignment. Now, both the reducer and the component care about the same part of the application domain and thus same or similar naming should be a safe bet here. Or, the other way around, you cannot rename one and leave the other in the current state of the code.

onboarding-task/test/models/ViewItem.test.js
```diff
@@ -0,0 +1,30 @@
...
+    const mockDataObject = { text: 'Mlock' };
+    const mockFlagsObject = { isBeingEdited: false };
+    const expectedViewItem = {
+      id: mockId,
+      index: 2,
+      text: 'Mlock',
+      isBeingEdited: false,
+    };
+
+    const createdViewItem = ViewItem(mockId, 2, mockDataObject, mockFlagsObject);
```
Suzii: Forgotten `new` keyword.

onboarding-task/src/containers/ListItem.js
```diff
@@ -1,31 +1,27 @@
...
-import * as ActionCreators from '../actions/actionCreators';
+import * as actionCreators from '../actions/actionCreators';
 import { ViewItem } from '../models/ViewItem';
 
-const createItemRecordMemoized = memoize(({ id, index, text, isEdited }) => (
-  new ViewItem({ id, index, text, isEdited })));
-
 const mapStateToProps = (state, { id, index }) => ({
-  item: createItemRecordMemoized({
+  item: new ViewItem(
```
Suzii: I liked the previous naming and usage a bit better, coz I knew straight away it is memoized. With this usage (the `new` keyword) I was not sure it would work before I really run the unit tests. However, it is an interesting solution :)

onboarding-task/test/actions/actionCreators.test.js
```diff
@@ -0,0 +1,74 @@
+import * as actionCreators from '../../src/actions/actionCreators';
+import * as actionTypes from '../../src/actions/actionTypes';
+
+describe('Action creators', () => {
+  it('creates "ITEM_CREATED" action correctly', () => {
```
Suzii: How about using string interpolation to get the actionCreator's name? Would be still up to date when you decide to change its name :wink:

onboarding-task/src/actions/actionCreators.js
```diff
@@ -1,33 +1,34 @@
-import * as ActionTypes from './actionTypes';
 import uuidV4 from 'uuid';
 
-export const createItem = (text) => ({
-  type: ActionTypes.ITEM_CREATED,
+import * as actionTypes from './actionTypes';
+
+export const createItem = (text, idGenerator = uuidV4) => ({
```
Suzii: You should provide a separate factory function and then call it with `uuidV4`. This factory will return the action creator itself, that would only accept the text as its argument. You should export both factory (used in tests) and the actionCreator (used in the container) separately. (Can be placed in one file for now, however, if it was more complex, the factory should be in its own file.)

onboarding-task/src/components/ListItem.jsx
```diff
@@ -1,47 +1,40 @@
...
+      onSave={props.onSave}
+      onCancel={props.onCancel}
+    /> :
+    <ViewItem
+      item={props.item}
+      onClick={props.onClick}
+    />;
+
+  return (
+    <span>
```
milan009: I believe that my intention here was to make the code more clear by not cluttering the return statement itself. It feels easier to follow when the decision *what* should be rendered is made before returning. Will the extra step of using the `item` variable (which I should change to `const`) have any noticeable impact on performance?

Suzii: No, I do not think it has any impact on the performance. However, I think `<span>` is completely redundant and if you do not want to return ternary, use `if`, `else` and return directly the JSX code.

milan009: That is a fair point, I will use `if` to return the code directly then. Thanks :)

onboarding-task/test/reducers/list/itemReducer.test.js
```diff
@@ -33,18 +21,24 @@ describe('Item reducer', () => {
   });
 
   describe('unknown action type', () => {
+    const action = { type: 'unknown' };
```
Suzii: You could have named this one `unknownAction` for sake of clarity.

PR #62, opened 2017-07-14
Feature/task 1

medvedovic:
Done


onboarding-task/src/components/ListItem.jsx
```diff
@@ -0,0 +1,58 @@
...
+      itemName: value,
+    });
+  };
+
+  _handleDeleteItemClick = () => {
+    this.props.onDeleteItem(this.props.id);
+  };
+
+  _toggleBeingEdited = () => {
+    this.setState({
```
Suzii: This should have been done via functional setState. Read more here: https://medium.com/@shopsifter/using-a-function-in-setstate-instead-of-an-object-1f5cfd6e55d1

onboarding-task/src/components/ListItem.jsx
```diff
@@ -0,0 +1,58 @@
...
+    itemName: PropTypes.string,
+    onDeleteItem: PropTypes.func,
+    onUpdateItem: PropTypes.func,
+  };
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      isBeingEdited: false,
+      itemName: this.props.itemName,
```
Suzii: There is no need to duplicate the item value in state in this component.

onboarding-task/src/components/ListItemInput.jsx
```diff
@@ -15,14 +16,17 @@ export class ListItemInput extends React.Component {
 
   componentDidMount() {
```
Suzii: As for this simple input form I think you can easily handle the 'Enter' scenario by wrapping the input in `<form>` tag and handling it in `onSubmit` callback. If you set the `button.type` to `submit` then you do not even separately handle the `button.onClick`.

onboarding-task/src/components/ListItemInput.jsx
```diff
@@ -15,14 +16,17 @@ export class ListItemInput extends React.Component {
...
     document.addEventListener('keydown', (event) => {
-      if ((event.keyCode === 13) && this.state.value) {
+      if ((event.keyCode === 13) && this.state.value && document.activeElement === this.textInput) {
```
Suzii: Also please handle validation by a separate function in separate file. Something like `isEmptyOrWhitespace` function would do 😉

onboarding-task/src/components/ListItemEditor.jsx
```diff
@@ -16,10 +17,31 @@ export class ListItemEditor extends React.Component {
...
+    document.addEventListener('keydown', (event) => {
+      if ((event.keyCode === 13) && this.state.itemName && document.activeElement === this.textInput) {
+        this._handleUpdateItemNameClick();
+      }
+    });
+
+    this._focus();
+  }
+
+  _focus = () => {
```
Suzii: Use `autofocus ` attribute for `input` element instead.

onboarding-task/src/components/ListItemEditor.jsx
```diff
@@ -16,10 +17,31 @@ export class ListItemEditor extends React.Component {
     };
   }
 
+  componentDidMount() {
+    // Cancel editation on ESC
+    document.addEventListener('keydown', e => {
```
Suzii: Two event listeners for `keydown` event - why aren't they handled in one function? Use switch inside.

onboarding-task/src/App.jsx
```diff
@@ -1,8 +1,39 @@
...
+    this.setState(prevState => ({
+      items: prevState.items.concat({ id: generateGUID(), itemName: value }),
+    }));
+  };
+
+  _updateItem = (key, value) => {
+    const newArray = [];
+    this.state.items.forEach(item => {
+      if (item.id === key) {
+        newArray.push({ id: key, itemName: value });
```
Suzii: This is correct, but not very clever way to handle item update. You are relying on the fact that the item contains exactly these two properties. If you decide to add some more in the future, this function will still be handling the update of the `itemName`. However, with this implementation you are loosing all other properties that might be present in the item object. Please, rewrite this code to make the item object extensible in the future without need for changing the already existing functionality -> use object spread operator (preferred) or `Object.assign()` function. If you are unsure about the differences or how to use them, feel free to ask :)

onboarding-task/src/utils/isTextInputValid.js
```diff
@@ -0,0 +1,3 @@
+export function isTextInputValid(input) {
+  return (input && input.trim().length > 0) === true;
```
Suzii: How about just
```
export const isNotEmpty = (value) => value && value.trim().length;
```

onboarding-task/src/components/List.jsx
```diff
@@ -29,20 +30,27 @@ export class List extends React.Component {
     const newArray = [];
     this.state.items.forEach(item => {
       if (item.id === key) {
-        newArray.push({ id: key, itemName: value });
+        newArray.push({
```
Suzii: You are still ignoring possible other properties of the item object.

onboarding-task/src/components/ListItem.jsx
```diff
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import { ListItemEditor } from './ListItemEditor';
 
 export class ListItem extends React.Component {
+  static displayName = 'displayName';
```
Suzii: No change... I am not sure whether you are testing my attention or patience or you just missed this one again.

onboarding-task/src/components/ListItemEditor.jsx
```diff
@@ -18,35 +20,42 @@ export class ListItemEditor extends React.Component {
...
-        this.props.onItemCancelEdit();
-      }
-    });
-    // Save on enter
-    document.addEventListener('keydown', (event) => {
-      if ((event.keyCode === 13) && this.state.itemName && document.activeElement === this.textInput) {
-        this._handleUpdateItemNameClick();
-      }
-    });
+    document.addEventListener('keydown', this._handleKeyboardInput);
```
Suzii: You can even add the `onKeyDown` event directly on the `input` element. Thet would simplify things a bit ;)

onboarding-task/test/isTextInputValid.test.js
```diff
@@ -0,0 +1,16 @@
+import { isTextInputValid } from '../src/utils/isTextInputValid';
+
+describe('isTextInputValid', () => {
```
Suzii: You are not testing whitespace characters, even though they are a special case in your implementation.

onboarding-task/src/components/ListItem.jsx
```diff
@@ -7,7 +7,7 @@ export class ListItem extends React.PureComponent {
   static displayName = 'displayName';
   static propTypes = {
     id: PropTypes.string.isRequired,
-    itemName: PropTypes.string.isRequired,
+    item: PropTypes.object.isRequired,
```
Suzii: You can use 
```
PropTypes.shape({
   id: PropTypes.string.isRequired,
   name: PropTypes.string.isRequired,
}).isRequired;
```

PR #66, opened 2017-07-27
Feature/task 4

Neytus:


onboarding-task/src/components/Node.tsx
```diff
@@ -30,29 +32,21 @@ const nodePropTypes = {
...
-      />
-    ) : (
-      <ViewNode
-        nodeViewModel={nodeViewModel}
-        onEdit={onEdit}
-      />
-    );
-  };
+export const Node: React.StatelessComponent<INodeProps> = props => {
+  return props.nodeViewModel.isBeingEdited ? (
```
PetrSvirak: OK, now this can be shortened even further – there is no need for block scope and return statement (this applies every time `return` statement is the first statement in a method's block :wink:)

onboarding-task/test/utils/validation.test.js
```diff
@@ -3,23 +3,38 @@ import { isNullOrWhitespace } from '../../src/utils/validation.ts';
 describe('validation', () => {
   describe('isNullOrWhitespace', () => {
     it('returns true on null input', () => {
-      expect(isNullOrWhitespace(null)).toBe(true);
+      const expectedValue = true;
+      const actualValue = isNullOrWhitespace(null);
+
+      expect(actualValue).toBe(expectedValue);
```
PetrSvirak: My bad I did not point this one earlier, sorry, however, there are [`toBeTruthy()`](https://facebook.github.io/jest/docs/expect.html#tobetruthy) and [`toBeFalsy`](https://facebook.github.io/jest/docs/expect.html#tobefalsy) method that would spare you an extra `expectedValue` declaration.

onboarding-task/test/utils/validation.test.js
```diff
@@ -3,23 +3,38 @@ import { isNullOrWhitespace } from '../../src/utils/validation.ts';
 describe('validation', () => {
   describe('isNullOrWhitespace', () => {
     it('returns true on null input', () => {
-      expect(isNullOrWhitespace(null)).toBe(true);
+      const expectedValue = true;
```
PetrSvirak: There is a line missing between arrangement and assertion parts of the test.

onboarding-task/src/actions/actionCreators.ts
```diff
@@ -0,0 +1,40 @@
+import {
+  TOGGLE_NODE,
+  SAVE_NODE,
+  DELETE_NODE,
+} from './actionTypes';
+import { addNodeFactory } from './addNodeFactory';
+import { generateId } from '../utils/generateId';
+
+export interface IAction {
```
PetrSvirak: Aha, I see why I mislead you here. The information in brackets referred to where the `IAction` interface is used, not where it should be placed. It should be placed in a standalone file → `.\actions\IAction.ts`.

onboarding-task/src/utils/generateId.ts
```diff
@@ -0,0 +1,17 @@
...
+  return sGuid;
+};
+
+const generateId = (): IdType => (
+  generateIdPart() + generateIdPart() + '-'
+  + generateIdPart() + '-'
+  + generateIdPart() + '-'
+  + generateIdPart() + '-'
+  + generateIdPart() + generateIdPart() + generateIdPart()
+);
```
PetrSvirak: I don't believe you need the outermost brackets here. Removing them will save you a line.

PetrSvirak:
We are almost there. Just few little details and the task is gonna be done :)

onboarding-task/src/components/AddNode.tsx
```diff
@@ -37,26 +40,32 @@ export class AddNode extends React.PureComponent<IAddNodeCallbacksProps, IAddNod
   };
 
   render() {
+    const keyHandlers: Partial<KeyHandler> = {
```
PetrSvirak: * I would preferably move the `Partial<T>` into the definition of `KeyHandler` – there is quite a big chance that not all commands will be mapped the bigger the app grows (and you actually experience it already).
* Also, move it out of the render.
* Simply pass `this.odAdd to the method` as the value of `saveNode`
* And, the `'` should not be necessary

onboarding-task/src/@types/global.d.ts
```diff
@@ -9,6 +9,15 @@ declare global {
...
   }
+
+  interface IKeyMap {
+    cancelNode: KeyCommands;
+    saveNode: KeyCommands;
+  }
+
+  type KeyCommands = 'esc' | 'enter';
+
+  type KeyHandler = {[key in keyof IKeyMap]: Function};
```
PetrSvirak: It seems that my comment towards _IAction_ got lost in translation. Please find it out. The bottom line is to move the `IAction` to a standalone file.  Same applies to the _IKeyMap_ and _KeyHandler_ (and _KeyCommands_ to be complete) – move these into a standalone file as well since they do not bring universal functionality or add (or change) types of a package library.

onboarding-task/src/components/EditableNode.tsx
```diff
@@ -64,8 +58,13 @@ export class EditableNode extends React.PureComponent<IEditableNodeProps, IEdita
   render() {
     const {text} = this.state;
 
+    const keyHandlers: Partial<KeyHandler> = {
```
PetrSvirak: Just a note: Partial should not be necessary here, should it?

PetrSvirak:
There are some details I mentioned in a "task-5" issue and will be handled later. They are related to duplicate usage of `isNullOrWhitespace` and key handlers stored in state.

PR #70, opened 2017-08-14
Feature/task 4

medvedovic:


medvedovic:
`keyMap` - after a long reevaluation, it probably would only complicate things or maybe it's easier than I think and I'm complicating it

`itemFactory` - it served a purpose when Peter was enlightening me in DI and other stuff, I am aware there is no need for such an over-complicated piece of code in application this small, however, there was a reason for that 

`data` - ... I renamed it from 'items' on Peter's request, not renaming it again

type aliases for id - considered, not doing it :D

The last paragraph took me a while to fully comprehend it and even more time to code it, hopefully, it's what you wanted, check `TypedRecord.ts` in models. 
However, it raises a question, is it possible to force the "inner class" to implement the interface that is passed in as TProperties?

Suzii:
Thanks for the changes, I love how you tackled the `TypedRecord` task :) There is no way (AFAIK) to force the inner class to implement TProperties. 

- as for the `itemFactoryBuilder` - yes, DI is super-important, but IMHO the builder does not bring you basically anything - it has its dependencies injected but itself does not need to be injected to the action creator. Anyway, leave it as is if you want.
- `IItemViewModel` - properties should be readonly
- place an empty line after each case statement for better readability
- as for the keyMap the only thing I wanted you to do was to define an interface like this and then not using strings constants as properties. Also specify that the constant is of type KeyMap every time you use it. (You can define Partial<KeyMap> if necessary.)
```
interface KeyMap {
  readonly saveKey: string;
  readonly cancelKey: string;
  readonly focusNewItemKey: string;
}

export const keyMap: KeyMap = {
  saveKey: 'enter',
  cancelKey: 'escape',
  focusNewItemKey: 'option+n',
};
```

medvedovic:
Fixed, let's merge the bitch

onboarding-task/src/components/ListItemCreator.tsx
```diff
@@ -0,0 +1,122 @@
...
+  render(): JSX.Element {
+    const { value } = this.state;
+    const globalHandlers: Partial<IKeyMapHandlers> = {
+      focusNewItemKey: () => {
+        if (this.textInput) {
+          this.textInput.focus();
+        }
+      },
+    };
+    const handlers = {
```
Suzii: This constant is still sad about not having explicit type :(

PR #71, opened 2017-08-28
Feature/task 5

VlastimilM:


onboarding-task/src/components/List.tsx
```diff
@@ -3,39 +3,70 @@ import * as PropTypes from 'prop-types';
...
+  };
+
+  componentWillMount() {
+    this.props.fetchItems();
+  }
+
+  getContent = () => {
+    if (this.props.isFetching) {
+      return <Spinner />;
+    } else {
```
Suzii: You do not need else here. if you got to this point it means, that the first condition was not met as you have the return statement inside.

onboarding-task/src/actions/actionCreators.ts
```diff
@@ -53,3 +100,40 @@ export const updateItemText = (id: string, text: string): IAction => ({
...
+  );
+
+  const fetchedItems = items.reduce((accu, item) => accu.set(item.id, item), Immutable.Map<string, Item>());
+  const fetchedItemsOrderedIds = Immutable.List<string>(items.map(item => item.id));
+
+  return {
+    type: FETCH_ITEMS_SUCCESS,
+    payload: {
+      items: fetchedItems,
+      orderedIds: fetchedItemsOrderedIds,
```
Suzii: I do not think this should be here at all. Reducer should do this job of extracting the ids. Your action creator should not be bothered by the store structure, it should only inform what has happend and pass the minimum amount of data to reducer so that it can do its job. This data is inferred, which means, reducer can easily compute it on its own from `items` property,.

onboarding-task/src/reducers/items/errorsReducer.ts
```diff
@@ -0,0 +1,24 @@
...
+  POST_ITEM_FAILURE,
+  DELETE_ERROR_MESSAGE,
+} from '../../constants/actionTypes';
+import { Error } from '../../models/Error';
+
+export type Errors = Immutable.List<Error>;
+
+export function errorsReducer(errors: Errors = Immutable.List<Error>(), action: IAction): Errors {
+  switch (action.type) {
+    case FETCH_ITEMS_FAILURE:
```
Suzii: The first two cases are same, merge them.

onboarding-task/src/components/ErrorAlert.tsx
```diff
@@ -0,0 +1,31 @@
...
+export interface IErrorAlertCallbacksProps {
+  onErrorDismiss: (errorId: string) => void;
+}
+export const ErrorAlert: React.StatelessComponent<IErrorAlertCallbacksProps & IErrorAlertDataProps> = ({ onErrorDismiss, error }) =>
+  (
+    <div className="alert alert-danger alert-dismissible" role="alert">
+      <button type="button"
+              className="close"
+              aria-label="Close"
+              onClick={() =>  onErrorDismiss(error.errorId)}
```
Suzii: Pass id of the error from container's ownProps.

PR #72, opened 2017-10-06
Feature/task 5

Neytus:


onboarding-task/src/models/NodeInfo.ts
```diff
@@ -2,14 +2,17 @@ import { TypedRecord } from './TypedRecord';
 
 interface INodeInfo {
```
vit-svoboda: `Info` is another of those misleading words that seem to mean something. NodeStatus might be a bit better.

onboarding-task/src/models/NodeContent.ts
```diff
@@ -1,7 +1,14 @@
 import { TypedRecord } from './TypedRecord';
 
-interface INodeContent {
-  readonly id: IdType;
+export interface IServerNode {
```
vit-svoboda: Not sure if the server model should be here. It looks like it's a superset of `INodeContent` while it can be completely different. I'd place it where you make the conversion from the server models to your client models.

onboarding-task/src/reducers/nodesListReducers/nodesInfoReducer.ts
```diff
@@ -0,0 +1,54 @@
...
+  POST_NODE_OPTIMISTIC,
+  PUT_NODE_SUCCESS,
+} from '../../actions/actionTypes';
+import { NodeInfo } from '../../models/NodeInfo';
+import { IAction } from '../../actions/IAction';
+import { INodeContent } from '../../models/NodeContent';
+
+export type INodesInfo = OrderedMap<Guid, NodeInfo>;
+
+export const nodesInfoReducer = (state: INodesInfo = OrderedMap<Guid, NodeInfo>(), action: IAction): INodesInfo => {
```
vit-svoboda: Same what I said about the `INodeInfo`. However, maybe I'd break this into two separate reducers: a set of IDs of edited nodes and the same with persisted nodes. There's no point trying to aggregate things.

onboarding-task/test/actions/actionCreators.test.js
```diff
@@ -1,68 +1,132 @@
 import * as actions from '../../src/actions/actionCreators.ts';
 import * as types from '../../src/actions/actionTypes.ts';
-import { addNodeFactory } from '../../src/actions/addNodeFactory.ts';
 
-const generateNewId = jest.fn(() => '80149842-a624-b66b-5d3c-37c24523ba46');
-const addNode = addNodeFactory(generateNewId);
+const id = '80149842-a624-b66b-5d3c-37c24523ba46';
 
 describe('actionCreators', () => {
```
vit-svoboda: You don't really have to test action creators. It's plenty if you test them together with the reducers. You can keep these for now, though ;)

onboarding-task/src/@types/global.d.ts
```diff
@@ -1,15 +1,15 @@
-import { IAction } from '../actions/IAction';
+import { IAppState } from '../reducers/IAppState';
 
-export type ThunkPromise = (dispatch: Dispatch, getState: any) => Promise<IAction>;
+export type IAction = {
+  type: string;
+  payload?: any;
+};
```
vit-svoboda: I don't think the contents of global.d.ts file are supposed to be exported. You just declare them globally and then you don't need to import them where you need them. Not a very nice solution, but can make things easier. However, if you want to export some interface so you can import it elsewhere, you should place it in another file.

onboarding-task/test/actions/actionCreators.test.js
```diff
@@ -1,4 +1,4 @@
-import * as actions from '../../src/actions/internalActionCreators/baseActionCreators.ts';
```
vit-svoboda: I'm pretty sure the rule about not importing `internalActionCreators` outside of the `actions` folder does not apply to tests. On the other hand, `publicActionCreators` should export only the actions that are supposed to be dispatched in the rest of the application. That's not true for those `stuffStarted` or `stuffFinished` actions. Those are just parts of a thunk the application dispatches.

onboarding-task/src/repositories/apiAdapter.ts
```diff
@@ -1,5 +1,5 @@
-import { API_URL } from '../../constants/urls';
-import { INodeContent } from '../../models/NodeContent';
+import { API_URL } from '../constants/urls';
+import { INodeContent } from '../models/NodeContent';
 
 export const checkStatus = (response: Response) => {
```
vit-svoboda: You're using this function only in this file, so there's no need to export it. You'd need to export it if you were to write tests for the method, though.

PR #75, opened 2017-10-13
[KA-200] Feature/task 1

matus12:


src/components/ItemList.jsx
```diff
@@ -0,0 +1,69 @@
...
+    return s4() + s4() + '-' + s4() + '-' + s4() + '-' +
+      s4() + '-' + s4() + s4() + s4();
+  }
+  handleChange(event) {
+    this.setState({
+      inputValue: event.target.value,
+    });
+  }
+  handleAdd() {
+    const newValues = [...this.state.values];
```
vit-svoboda: You're creating the `newValues` even if you never need them (due to the inputValue being empty). If you created this array later, you could inline the addition like this: `[...this.state.values, [this.state.input, false]`.

src/components/ItemList.jsx
```diff
@@ -0,0 +1,69 @@
...
+    newValues[index][1] = editable;
+    this.setState({
+      values: newValues,
+    });
+  }
+  render() {
+    return (
+      <ul className="list-group">
+        {this.state.values.map((val, index) => <ListItem key={this.guid()} editable={val[1]} text={val[0]} actionDelete={() => this.handleDelete(index)} actionEdit={(text) => this.handleEditText(index, text)} handleEditableState={(editable) => this.handleEditableState(index, editable)} />)}
```
vit-svoboda: You're creating a lot of stuff (mostly functions) here, in a `render` method. I hear it's frowned upon. A better way would be to have reusable functions in your component, i.e. instead of `actionDelete={() => this.handleDelete(index)}` you would just `actionDelete={this.handleDelete}`. That way you need to make it pass the parameter, though.
Also, generating the key like this seems like a bit of an overkill. Is the index not unique enough for react to render the children? And if you already come to a situation where you need the item to have some sort of an ID, how about adding an id property to the item and using it instead of the indices? I think it would be a bit better :)
And lastly, the line is long af ;)

src/components/ItemList.jsx
```diff
@@ -0,0 +1,69 @@
...
+        <li className="list-group-item">
+          <div className="col-xs-4">
+            <input className="form-control" value={this.state.inputValue} onChange={(event) => this.handleChange(event)} />
+          </div>
+          <button type="button" className="btn btn-light" onClick={() => this.handleAdd()}>Add</button>
+        </li>
```
vit-svoboda: This add form could be a separate component.

src/components/ListItem.jsx
```diff
@@ -0,0 +1,51 @@
+import React, { PureComponent } from 'react';
+
+export class ListItem extends PureComponent {
+  constructor(props) {
+    super(props);
+    this.state = {
+      editable: props.editable,
+      text: props.text,
```
vit-svoboda: Do you need to copy the `props` to `state`? The component has access to `props` everywhere I can think of.

src/components/ListItem.jsx
```diff
@@ -0,0 +1,51 @@
...
+      text2: props.text,
```
vit-svoboda: `text2` is rather poor property name. What is it? The edited text? Or a backup in case of an edit?

src/components/ListItem.jsx
```diff
@@ -0,0 +1,51 @@
...
+        {(this.state.editable) ?
+          <div>
+            <div className="col-xs-4">
+              <input className="form-control" value={this.state.text} onChange={(event) => this.handleChange(event)} />
+            </div>
+            <button type="button" className="btn btn-primary" onClick={() => this.handleSave()}>Save</button>
+            <button type="button" className="btn btn-light" onClick={() => this.handleCancel()}>Cancel</button>
+            <button type="button" className="btn btn-danger" onClick={() => this.props.actionDelete()}>Delete</button>
+          </div> :
+          <div onClick={() => this.handleClick()}>{this.state.text}</div>}
```
vit-svoboda: Just pass the functions you have, no need for these arrow functions.

src/components/ListItem.jsx
```diff
@@ -0,0 +1,51 @@
...
+
+  handleSave() {
+    this.setState({
+      text2: this.state.text,
+      editable: false,
+    });
+    this.props.actionEdit(this.state.text);
+    this.props.handleEditableState(false);
+  }
+  handleCancel() {
```
vit-svoboda: The empty spaces between functions should be consistent. I'd say 1 empty line is a good choice.
Don't worry about the unnecessary text you're sending to the clients, you'll minify the scripts for production anyway.

src/components/ListItem.jsx
```diff
@@ -0,0 +1,51 @@
...
+    };
+  }
+
+  handleSave() {
+    this.setState({
+      text2: this.state.text,
+      editable: false,
+    });
+    this.props.actionEdit(this.state.text);
+    this.props.handleEditableState(false);
```
vit-svoboda: The `actionEdit` could also handle the state. If it's what I think it is. Perhaps something like `onItemEdited` would be a better name.

src/components/List.jsx
```diff
@@ -13,21 +21,9 @@ export class List extends PureComponent {
...
         </div>
-
-        <div className="row">
-          <div className="col-sm-12">
-            <p className="lead text-center">Desired functionality is captured in the gif image. </p>
-            <p className="lead text-center"><b>Note: </b>Try to make solution easily extensible (e.g. more displayed fields per item like
-              <code>dateCreated</code>).</p>
-            <img src={assignment} alt="assignment" className="img--assignment" />
-          </div>
-        </div>
```
vit-svoboda: rip gif, though you still import the image.

src/components/ItemList.jsx
```diff
@@ -0,0 +1,69 @@
...
+  guid() {
+    function s4() {
+      return Math.floor((1 + Math.random()) * 0x10000)
+        .toString(16)
+        .substring(1);
+    }
+    return s4() + s4() + '-' + s4() + '-' + s4() + '-' +
+      s4() + '-' + s4() + s4() + s4();
+  }
+  handleChange(event) {
```
vit-svoboda: `handle` seems like a useful word, but does it really say much about what's going on?

src/components/ListItem.jsx
```diff
@@ -0,0 +1,51 @@
...
+      <li className="list-group-item">
+        {(this.state.editable) ?
+          <div>
+            <div className="col-xs-4">
+              <input className="form-control" value={this.state.text} onChange={(event) => this.handleChange(event)} />
+            </div>
+            <button type="button" className="btn btn-primary" onClick={() => this.handleSave()}>Save</button>
+            <button type="button" className="btn btn-light" onClick={() => this.handleCancel()}>Cancel</button>
+            <button type="button" className="btn btn-danger" onClick={() => this.props.actionDelete()}>Delete</button>
+          </div> :
```
vit-svoboda: If you wanted to be super hardcore, you could extract also this item editor to a separate component.

src/components/ItemList.jsx
```diff
@@ -0,0 +1,69 @@
+import React, { PureComponent } from 'react';
+
+import { ListItem } from './ListItem';
+
+export class ItemList extends PureComponent {
```
vit-svoboda: If your component uses props, you should declare their `propTypes` so you have at least some control over what it's getting.

src/components/List.jsx
```diff
@@ -50,16 +69,19 @@ export class List extends PureComponent {
...
+                  index={index}
+                  onDeleteItem={this.deleteItem}
+                  onSaveItem={this.saveItem}
+                />,
+              )
+            }
             <AddItem
-              value={this.state.inputText}
-              onAddItem={this.onAddItem}
+              onAddItem={this.addItem}
```
PetrSvirak: This one can be on the same line as the `<AddItem` - it is OK to have single attribute assignment inlined.

src/components/PlainItem.jsx
```diff
@@ -0,0 +1,21 @@
+import React, { PureComponent } from 'react';
```
PetrSvirak: You renamed this component without losing history. Very nice :+1:

test/generateId.test.js
```diff
@@ -0,0 +1,36 @@
...
+
+describe('Generate id', () => {
+  it('generates unique id', () => {
+    const guid1 = generateId();
+    const guid2 = generateId();
+
+    expect(guid1).not.toEqual(guid2);
+  });
+
+  it('returns string of length 36', () => {
```
PetrSvirak: This test and the tests below are great (as a matter of fact, all of them are, thanks :+1: ) However, I propose to merge them into a single one with a regular expression describing the GUID – I guess you can google it or write a single regular expression ensuring the will look exactly as it should. These are, as I said, good as well though!

src/components/AddItem.jsx
```diff
@@ -0,0 +1,61 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+
+export class AddItem extends PureComponent {
+  static propTypes = {
+    onAddItem: PropTypes.func.isRequired,
+    chooseFormStyle: PropTypes.func.isRequired,
```
vit-svoboda: If you have this low-level component, it's only real responsibility is the markup. With props like `chooseFormStyle`, you take the responsibility away.
I would understand that if you didn't want to pass the `isInputValid` flag to it, but you do that anyway. So the component has all the information to chose appropriate class for the input. Setting a `className` for the whole component could be considered as a customization of the component, but here you change one class in a whole hierarchy of elements and classes making it fragile.

src/components/AddItem.jsx
```diff
@@ -0,0 +1,61 @@
...
+    isInputValid: PropTypes.func.isRequired,
+    fillInTitle: PropTypes.func.isRequired,
```
vit-svoboda: This, again, is sort of an awkward callback prop. It's not handling of some action that happens, it's not data for the component. You just pass some _utility_ function down from the parent component. I think the component already has all the information to make the decision. If you wanted to show different messages in different usages of the component, you could pass simple string prop `validationError` or something like that, that the component decides to show as a button tooltip in case it's disabled due to the value not being valid.
Ehh, sorry about the wall of text. Do you understand what I'm trying to say?

src/components/AddItem.jsx
```diff
@@ -0,0 +1,61 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+
+export class AddItem extends PureComponent {
+  static propTypes = {
+    onAddItem: PropTypes.func.isRequired,
+    chooseFormStyle: PropTypes.func.isRequired,
+    isInputValid: PropTypes.func.isRequired,
```
vit-svoboda: As you can see in the comments around, that make a bit less sense in the light of this - I expected the `isInputValid` to be a boolean saying whether the input is valid. But it's a function and it's a function for a good reason, as the parent component doesn't know the input value. However, I'm not sure this passing of business logic down the children is a good idea. You could either move this function to some utils file and import it everywhere you need it. Or you can create a container that will take the responsibility for state handling of the value and also for the validation.

src/components/AddItem.jsx
```diff
@@ -33,10 +33,9 @@ export class AddItem extends PureComponent {
       <li className="list-group-item">
         <div className="col-xs-4">
           <input
-            className="form-control"
+            className={isInputValid ? 'form-control' : 'form-control2'}
```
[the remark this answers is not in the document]
matus12: Renamed, bootstrap has in docs 'Input with error' which is input with red border, but it is red all the time, so when u open the app there is red bordered input with add button which for me doesn't seem nice. That's why I created my own and it is red only when u have focus on that input.

test/inputValidation.test.js
```diff
@@ -0,0 +1,15 @@
...
+const emptyInput = '';
+const validInput = 'asdfasf';
+const invalidInput = '          ';
+
+describe('Input validation', () => {
+  it('validates input correctly', () => {
+    expect(isInputValid(validInput)).toEqual(true);
+    expect(isInputValid(emptyInput)).toBe(false);
+    expect(isInputValid(invalidInput)).toBe(false);
+  });
```
PetrSvirak: This is not very nice.
* Each of your tests should have exactly one `expect`. You can break this rule in order to validate multiple aspects of tested app state, but not to merge tests together.
* Tests should consist of 3 parts - arrange and act and assert (i.e. [AAA](http://wiki.c2.com/?ArrangeActAssert)) – it makes tests more readable, and in your case, `expect`s shorter. 
* Your test descript (in `it`) it too vague and fixture name (`describe`) a bit misleading, merging different tests together is the cause. Let's use tested method name in the `describe` and then make each test's description worth it - e.g. (_"isInputValid"_)  _"returns true for non-empty string"_ could be the first case.

I know this will make the file longer, but each individual test will be more readable.

src/utils/inputValidation.js
```diff
@@ -0,0 +1,2 @@
+export const isInputValid = (inputText) =>
```
PetrSvirak: You should not name functions (or classes or instances) based on a place they are used at, rather name them by their features and properties. When you describe a knife, you won't probably use "killers' tool" but rather "tool with a handle and sharp edge" (more on-topic example: this function could be used to validate data coming in from a file. While it is not totally off-track, input there would be a file and only certain part of it (1 item out of thousands) could have empty text).

src/components/AddedItem.jsx
```diff
@@ -0,0 +1,60 @@
...
+    this.state = ({
+      inputText: '',
+      isInputValid: false,
+    });
+  }
+
+  changeOfInput = (event) => {
+    this.setState({
+      inputText: event.target.value,
+      isInputValid: validateText(event.currentTarget.value),
```
PetrSvirak: First, you mix `currentTarget` and `target`  - while they are same in this case, it is misleading (see this [SO]())

PetrSvirak: Second, read through [this article](https://medium.learnreact.com/setstate-takes-a-function-56eb940f84b6) – you see the the bug in your code then and fix it 😉

src/components/AddedItem.jsx
```diff
@@ -0,0 +1,60 @@
...
+      isInputValid: false,
+    }));
+  };
+
+  render() {
+    return (
+      <li className="list-group-item">
+        <div className="col-xs-4">
+          <Input
+            text={this.state.inputText}
```
PetrSvirak: In the native `input` component it is called `value` I vote for preserving the name.

src/components/AddedItem.jsx
```diff
@@ -0,0 +1,60 @@
...
+            isInputValid={this.state.isInputValid}
+            onChange={this.changeOfInput}
+          />
+        </div>
+        <button
+          type="button"
+          title={(this.state.isInputValid)
```
PetrSvirak: Please, extract this assignment into a standalone variable. 
Pro tip: Use `CTRL` + `ALT` + `V` to get the job done.

src/components/Input.jsx
```diff
@@ -0,0 +1,56 @@
...
+          })
+        }
+      >
+        <input
+          className="form-control"
+          value={this.props.text}
+          onChange={this.props.onChange}
+          onFocus={this.focus}
+          onBlur={this.blur}
+          title={(this.props.isInputValid)
```
PetrSvirak: This is the 3rd occurrence of the code. Unfortunately, this one renders the code less reusable. What about passing in the error message itself? If it is undefined, the `value` is valid, if there is a message, `value` is undefined. This way, your component will be both more reusable and few lines shorter 😃

src/components/Input.jsx
```diff
@@ -0,0 +1,56 @@
+import React, { PureComponent } from 'react';
+import classnames from 'classnames';
+import PropTypes from 'prop-types';
+
+export class Input extends PureComponent {
+  static propTypes = {
+    onChange: PropTypes.func.isRequired,
+    text: PropTypes.string.isRequired,
+    isInputValid: PropTypes.bool.isRequired,
```
PetrSvirak: You refer to `text` (or `value` more precisely) here. Anyway, it seems clear enough to me in this context to call the prop simply `isValid`.

src/components/Input.jsx
PetrSvirak: I will change the assignment here a bit for you to demonstrate why this is a problem: I want the `AddedItem` component to state that _"New item must consist of non-empty text"_ and the `EditedItem` component to state that _"Empty item cannot be stored. \n Tip: Use delete button to remove an item"_ (while changing the message on 3 places might be boring enough to prove my point, I believe this is a bit funnier to code 😇)

src/components/AddedItem.jsx
PetrSvirak: I happened to think you are using function `setState` - sorry for that. There is no bug in your code (or at least not the one I thought it was in the second comment) :upside_down_face:

PetrSvirak:
Job well done, thanks 👍

PR #76, opened 2017-10-16
Feature/task 1

kubmir:


src/components/AddLine.jsx
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  updateInsertedText = (event) => {
+    this.setState({
+      insertedText: event.target.value,
+    });
+  }
+
+  handleEnterPress = (event) => {
+    if (event.key === 'Enter') {
```
PetrSvirak: First of all: :+1: I like this fine tough.

You should really think about the app in global though. Why not using hotkeys in a unified way - for example `Enter` does not work on editing and I would expect `Esc` to equal clicking on the `Cancel` button (and maybe even `Delete` removing the member). Take this as highly optional, though, if you are interested, there is quite [a nice package](https://www.npmjs.com/package/react-hotkeys) to help you out. Use the package (can be in one of the following tasks or something; that's optional) and remove the event handler completely (within this task). 

Side notes: 
* I don't want to push you into given package, however, there should definitely be a unified solution for all your inputs within the app. 
* Instead in the package, the solution here might, for example, lay in interconnection with validation (see below) and it might be a single reusable `input` component, however, you might come-up to some obstacles since you don't want buttons to be part of the component. 
* When using hot-keys, you will inevitably come into contact with element focusing – you need to focus the (latest) input when it is rendered for an existing member (react helps you out since it does not re-renders other inputs when clicking on particular (not-yet-edited) member) and you need to focus addition input when no existing member is being edited
* you also need to decide, for example, whether shortcuts should work in inputs only – here the simple answer is yes, because of multiple editable members… – in general, you need to think this through :angel:

src/components/NonEmptyInput.jsx
```diff
@@ -0,0 +1,79 @@
...
+  };
+
+  onExitingInput = () => {
+    if (this.props.checkIsTouched !== undefined) {
+      this.props.checkIsTouched(false);
+    }
+  };
+
+  onInputChange = (e) => {
+    this.props.updateInsertedText(e);
```
vit-svoboda: You could pass just the target value to the props callback. So it doesn't have to deal with some event argument.

src/components/NonEmptyInput.jsx
```diff
@@ -0,0 +1,79 @@
...
+
+  componentDidMount() {
+    const length = this.textInput.value.length;
+    this.textInput.focus();
+    this.textInput.setSelectionRange(length, length);
+  }
+
+  cancelFocusOfInput = () => {
+    this.textInput.blur();
+  };
```
vit-svoboda: This works quite nice in the component for addition, but in the editor, if enter does the same as a "Save" click, one might expect escape key to behave as a "Cancel" click, not just to blur the input.

src/components/NonEmptyInput.jsx
```diff
@@ -0,0 +1,79 @@
...
+
+  onEnterClick = () => {
```
vit-svoboda: How does one _click_ the enter? :) Perhaps, I'd stick with the name of the operation (`'saveChanges'`) instead.

src/components/NonEmptyInput.jsx
```diff
@@ -0,0 +1,79 @@
...
+
+    return (
+      <div className={inputErrorStyle}>
+        <HotKeys handlers={handlers}>
+          <input
+            type="text"
+            ref={(input) => {
+              this.textInput = input;
+            }}
+            className="form-control"
```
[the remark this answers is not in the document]
kubmir: I added inputClassName as a prop of the NonEmptyInput component and I passed the value to this prop from parents. However, I read on the internet that has-error CSS class should be added to the parent element of input. I think it is not okay to slap both classes on the input element.

_Bootstrap includes validation styles for error, warning, and success states on form controls. To use, add .has-warning, .has-error, or .has-success to the parent element. Any .control-label, .form-control, and .help-block within that element will receive the validation styles._ - statement from https://getbootstrap.com/docs/3.3/css/ .

Is it acceptable now?

PR #77, opened 2017-10-19
Feature/task 5

medvedovic:


onboarding-task/src/actions/httpActionFactories/postDataActionFactory.ts
```diff
@@ -0,0 +1,45 @@
...
+} from '../../models/ItemDataDTO';
+import { IAction } from '../IAction';
+import { IItemDataActionDependencies } from './itemDataActionFactory';
+
+interface IPostItemDataActionFactoryDependencies extends IItemDataActionDependencies {
+  createItemOperation: (value: string) => IAction;
+}
+
+/**
+ * Core function used in both derived functions
```
vit-svoboda: Well... the functions are not derived per se. They may be using this _core_ but I still think the "core" part of the name should go away.

onboarding-task/src/components/ListItem.tsx
```diff
@@ -10,38 +10,42 @@ export interface IListItemDataProps {
 }
 
 export interface IListItemCallbacksProps {
-  onDeleteItem: () => void;
-  onUpdateItem: (value: string) => void;
+  onDeleteData: () => void;
+  onUpdateData: (value: string) => void;
```
vit-svoboda: What was wrong with deleting/updating items? To me, the generic "data" seems to be a step back.

onboarding-task/src/models/ListItemData.ts
```diff
@@ -8,11 +8,14 @@ interface IListItemData {
   readonly id: string;
   /** Value held by item */
   readonly value: string;
+  /** Id representation saved in store */
+  readonly localId: string;
```
vit-svoboda: What is the difference between the `id` and the `localId`? Are you not using the `id` at all anymore?

onboarding-task/src/constants/initialState.ts
```diff
@@ -1,46 +1,15 @@
...
-      [id2, new ListItemFlags()],
-      [id3, new ListItemFlags()],
-      [id4, new ListItemFlags()],
-    ]),
+    ids: List(),
+    data: OrderedMap(),
+    flags: Map(),
   },
+  app: {
+    list: new AppSettings,
```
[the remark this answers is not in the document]
medvedovic: ... fair enough

onboarding-task/src/reducers/app/app.ts
```diff
@@ -0,0 +1,6 @@
+import { combineReducers } from 'redux';
+import { settings } from './list/settings';
+
+export const app = combineReducers({
+  list: settings,
```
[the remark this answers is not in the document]
medvedovic: Not really necessary to wrap them, I agree, it was designed for easier extensibility just in case I needed more "settings" for these smaller apps

onboarding-task/src/constants/actionTypes.ts
```diff
@@ -1,4 +1,28 @@
...
+export namespace FetchData {
+  export const IS_LOADING = 'FETCH_IS_LOADING';
+  export const HAS_FAILED = 'FETCH_HAS_FAILED';
+  export const HAS_SUCCEEDED = 'FETCH_HAS_SUCCEEDED';
+}
+
+export namespace ItemActions {
+  export const POST_ITEM_TO_SERVER = 'POST_ITEM_TO_SERVER';
+  export const DELETE_ITEM_TO_SERVER = 'DELETE_ITEM_TO_SERVER';
+  export const PUT_ITEM_TO_SERVER = 'PUT_ITEM_TO_SERVER';
```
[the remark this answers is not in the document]
medvedovic: Nupe :(

onboarding-task/src/actions/httpActionFactories/fetchBuilder.ts
```diff
@@ -0,0 +1,18 @@
...
+      },
+      body: JSON.stringify(body)
+    };
+
+    return injectedFetch(url, requestParameters)
+      .then((response: Response) => {
+        if (!response.ok)
+          throw new Error(response.statusText + ' at ' + response.url);
+
+        return response;
```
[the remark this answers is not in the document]
medvedovic: I could not as it is not possible to call .json() when receiving response from delete

onboarding-task/src/actions/httpActionFactories/postDataActionFactory.ts
vit-svoboda: Actually, I think the comment should go as well. It's just misleading and doesn't help anybody.

onboarding-task/src/actions/httpActionFactories/fetchBuilder.ts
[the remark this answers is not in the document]
vit-svoboda: Right, sorry about that.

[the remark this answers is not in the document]
PetrSvirak: Well, that is sad right? Then you might actually leverage factory pattern - it is meant to create different instances based on given parameters. 
See this [SO](https://stackoverflow.com/questions/1541314/when-does-it-make-more-sense-to-use-the-factory-pattern-rather-than-an-overloade) first. Then, change your point of view a bit: 

You are trying to create a `Promise` and it might be the result of one of the following paths: You either call fetch and return it (delete; `: Promise`) or call `.json()` on it (other verbs; `: Promise<Object>`) or call `.text()` on it (you actually have no case when your API does not return (valid) JSON so I mention it for some sake of completeness and will no more; `: Promise<string>`).

Now, You could have a factory (or builder since JS does not distinguish between types too much) that (receives an enum or provides two methods and that) either returns `fetch`'s promise or `fetch.json`'s promise and then configure your dependencies properly instead of the current approach.

What I'm saying: It is definitely possible :wink:

onboarding-task/src/components/ListItem.tsx
```diff
@@ -13,35 +14,46 @@ export interface IListItemCallbacksProps {
...
+      method={itemViewModel.failedHttpAction}
       onClick={onToggleBeingEdited}
+      onResendRequest={onResendRequest}
+    />
+    :
+    <ListItemDisplayWithRedo
+      value={itemViewModel.value}
+      method={itemViewModel.failedHttpAction}
+      onResendRequest={onResendRequest}
     />
```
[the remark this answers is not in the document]
PetrSvirak: I mostly agree wit vit (couldn't help it, sorry 👼), but it would be nice if the first ternary did not have two 2 "else" parts on the first glance. Otherwise, it would not be so nice though.
Anyway, this should lead you (@medvedovic) to a possibility that these two flags actually describe one concept and thus should not be treated separately in a first place (of course one can describe any concept by a suitable array of true/false flags 🤣)

onboarding-task/src/actions/httpActionFactories/itemDataActionFactory.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export const itemDataActionFactory = (
+  core: (
+    dependencies: IItemDataActionDependencies,
+    dispatch: Dispatch,
+    url: string,
+    localId: string,
+    itemDto: IServerItemDataViewModel | undefined
+  ) => Promise<IAction>, dependencies: IItemDataActionDependencies) =>
```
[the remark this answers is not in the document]
medvedovic: Actually, I refactored the app so that the core is not there anymore :)

onboarding-task/src/components/App.tsx
```diff
@@ -0,0 +1,58 @@
...
+        </div>
+      }
+
+      <section className="shortcuts">
+        <h3>List of shortcuts</h3>
+        <ul>
+          <li><code>alt + n</code>Focus main input</li>
+          <li><code>escape</code>Cancel</li>
+        </ul>
+        <p>More to be added...</p>
```
vit-svoboda: I don't think you're gonna bother ;)

onboarding-task/src/index.js
```diff
@@ -18,7 +19,7 @@ const store = createStore(
   rootReducer,
   initialState,
   composeEnhancers(
-    applyMiddleware(logger)
+    applyMiddleware(logger, thunk)
```
vit-svoboda: Apparently the order of the middlewares matters and this way, you also get logged all dispatched thunks in the console in a less than perfect way, see https://github.com/gaearon/redux-thunk/issues/91

onboarding-task/src/constants/HttpAction.ts
```diff
@@ -0,0 +1,6 @@
+export enum HttpAction {
+  GET = 'GET',
+  POST = 'POST',
+  DELETE = 'DELETE',
+  PUT = 'PUT',
+}
```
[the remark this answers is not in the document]
medvedovic: The only difference afaik is that enum gets compiled to js as a function, while type does not, so it would probably help to save several bytes in compiled code? :D

onboarding-task/src/components/App.tsx
medvedovic: Actually, I have been working on this on my own computer at home, however, I still don't seem to be able to find some time to finish that

onboarding-task/src/constants/actionTypes.ts
```diff
@@ -1,4 +1,25 @@
...
+  export const SUCCESS = 'POST_ITEM_TO_SERVER_SUCCESS';
+  export const FAILURE = 'POST_ITEM_TO_SERVER_FAILURE'
+}
+
+export namespace PUT_ITEM_TO_SERVER {
+  export const SUCCESS = 'PUT_ITEM_TO_SERVER_SUCCESS';
+  export const FAILURE = 'PUT_ITEM_TO_SERVER_FAILURE';
+}
+
+export const DELETE_ITEM_TO_SERVER_FAILURE = 'DELETE_ITEM_TO_SERVER_FAILURE';
```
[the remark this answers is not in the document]
medvedovic: Let's pretend I've never written anything like this... must have been wind :smile:

onboarding-task/src/actions/httpActionFactories/putDataThunkFactory.ts
```diff
@@ -0,0 +1,95 @@
...
+    },
+  },
+});
+
+export const putData = putDataThunkFactory({
+  operation: fetchBuilder(fetch),
+  transformDataToDto: toServerItemDataViewModel,
+  onSuccess: handleSuccessfulRequest(PUT_ITEM_TO_SERVER.SUCCESS),
+  onError: handleErrorRequest(PUT_ITEM_TO_SERVER.FAILURE),
+  updateItem: updateItem,
```
vit-svoboda: Use shorthand property if the property and the name of the variable with value are the same. It's sexy :P

onboarding-task/src/actions/httpActionFactories/deleteItemThunkFactory.ts
```diff
@@ -0,0 +1,42 @@
...
+import { handleErrorRequest } from './requestStatusActions';
+import { DELETE_ITEM, DELETE_ITEM_AT_SERVER_FAILURE } from '../../constants/actionTypes';
+import { apiEndpoint } from '../../constants/apiEndpoint';
+import { fetchBuilder } from './fetchBuilder';
+
+export interface IDeleteItemThunkFactoryDependencies {
+  operation: (url: string, httpMethod: HttpAction) => Promise<Response>;
+  onSuccess: (_itemId: string, _response: Response) => IAction;
+  onError: (_itemId: string, _error: Error) => IAction;
+  httpMethod: HttpAction;
```
vit-svoboda: I don't think the method should be injected into the `deleteItemThunkFactory` - it seems the URL and the HTTP action should be its responsibility.

onboarding-task/src/actions/httpActionFactories/deleteItemThunkFactory.ts
```diff
@@ -0,0 +1,42 @@
...
+        .catch((error) =>
+          dispatch(dependencies.onError(itemId, error)));
+    };
+
+export const deleteItem = (id: string): IAction => ({
+  type: DELETE_ITEM,
+  payload: {
+    id,
+  },
+});
```
vit-svoboda: These basic action creators used to be in a separate file for a good reason. You use them here and in the reducer tests and you usually don't want to drag all the things you import here for the dependency injection to a test file. In this case, it's alright, but if you wanted to do e.g. something with the `localStorage` here, the tests would fail, as there's no `localStorage` in node.

onboarding-task/src/actions/httpActionFactories/deleteItemThunkFactory.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export const deleteData = deleteItemThunkFactory({
+  operation: fetchBuilder(fetch),
+  onError: handleErrorRequest(DELETE_ITEM_AT_SERVER_FAILURE),
+  onSuccess: deleteItem,
+  httpMethod: HttpAction.DELETE,
+  apiEndpoint
+});
```
vit-svoboda: I think having the thunk factory assembly in a separate file is better, mostly for the same reason I'd place the basic action creator elsewhere. You want to test the thunk without importing the whole of the dependencies.

onboarding-task/src/actions/httpActionFactories/deleteItemThunkFactory.ts
```diff
@@ -0,0 +1,42 @@
+import 'isomorphic-fetch';
+import { IAction } from '../IAction';
+import { HttpAction } from '../../constants/HttpAction';
+import { handleErrorRequest } from './requestStatusActions';
+import { DELETE_ITEM, DELETE_ITEM_AT_SERVER_FAILURE } from '../../constants/actionTypes';
+import { apiEndpoint } from '../../constants/apiEndpoint';
+import { fetchBuilder } from './fetchBuilder';
+
+export interface IDeleteItemThunkFactoryDependencies {
+  operation: (url: string, httpMethod: HttpAction) => Promise<Response>;
```
vit-svoboda: Maybe the `operation` has become too general a name for the property since now it's just an abstraction around an HTTP request that allows you not to worry about headers etc.

onboarding-task/src/actions/httpActionFactories/deleteItemThunkFactory.ts
```diff
@@ -0,0 +1,42 @@
+import 'isomorphic-fetch';
+import { IAction } from '../IAction';
+import { HttpAction } from '../../constants/HttpAction';
+import { handleErrorRequest } from './requestStatusActions';
+import { DELETE_ITEM, DELETE_ITEM_AT_SERVER_FAILURE } from '../../constants/actionTypes';
+import { apiEndpoint } from '../../constants/apiEndpoint';
+import { fetchBuilder } from './fetchBuilder';
+
+export interface IDeleteItemThunkFactoryDependencies {
```
vit-svoboda: You don't need to export this interface. The only other use is in the tests when you create the dependency object and type it to this. However, if you didn't type it at all, TypeScript will use its shape and compare it with the shape of the dependency parameter and if there was something missing, you'd get the same error you get now. It's just unnecessary.

onboarding-task/src/actions/httpActionFactories/requestStatusActions.ts
```diff
@@ -0,0 +1,37 @@
+import { IServerItemDataModel } from '../../models/IServerItemDataModel';
+import { IAction } from '../IAction';
+import { ListItemData } from '../../models/ListItemData';
+import { POST_ITEM_TO_SERVER } from '../../constants/actionTypes';
+
+export const handleSuccessfulPost = (temporaryId: string, params: IServerItemDataModel) => ({
```
vit-svoboda: These are just action creators, they do no _handling_. Furthermore, "handle" is a somewhat empty word. It looks important but doesn't really say anything. Just like "manage" and "provide". We don't like those ;)

onboarding-task/src/actions/httpActionFactories/deleteItemThunkFactory.ts
medvedovic: I really had no idea how to call that :rofl:

onboarding-task/src/actions/actionCreators.ts
```diff
@@ -1,12 +1,68 @@
...
+export const fetchHasSucceededBuilder = (factory: (value: string, id: string) => ListItemData) =>
+  (items: Array<IServerItemDataModel>) => ({
+    type: FETCH_DATA.HAS_SUCCEEDED,
     payload: {
-      item: factory(value),
+      items: items.map(item => factory(item.value, item.id))
+    }
+  });
+
+export const fetchHasSucceeded = fetchHasSucceededBuilder(listItemDataConverter);
```
vit-svoboda: I don't think the action creator should do any data transformation. You pass `items` in it and it wraps them in a payload with proper action type, that's it. The conversion from `IServerItemDataModel` to `ListItemData` should be performed in the `fetchDataThunkFactory` or in the appropriate reducer. The reducer would be probably cleaner, but on the other hand, your other thunks are already responsible for data transformation as you send to the server and back. So I'd move the `listItemDataConverter` to `fetchDataThunkFactory` dependencies and use it before the thunk calls `onFetchSucceeded`.

onboarding-task/test/actions/httpActionFactories/postDataThunkFactory.test.ts
```diff
@@ -0,0 +1,177 @@
...
+      type: CREATE_ITEM,
+      payload: {
+        item: new ListItemData({id, value: 'Go home'})
+      }
+    };
+    const postExpectedResult = {
+      type: POST_ITEM_TO_SERVER.SUCCESS,
+      payload: {id, value: 'Go home'}
+    };
+    const postItemAsync = postItemDataThunkFactory({ ...dependencies })('Go home');
```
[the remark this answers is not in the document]
medvedovic: To be honest with you, I did not realize what I was doing, now that I do, I feel ashamed

onboarding-task/src/actions/actionCreators.ts
medvedovic: Good call

PR #79, opened 2017-10-30
[Task 1] Editable list

LindaLangerova:


src/components/ListItem.jsx
```diff
@@ -0,0 +1,98 @@
...
+  };
+
+  renderNotEditedItem = () => {
+    const { item } = this.props;
+    return (
+      <div
+        className="list-group-item"
+        id="to-be-counted"
+        onClick={this.toggleTextEditing}
+      >{item.value}</div>
```
vit-svoboda: The preferred formatting of JSX is like this:
```
<div
  className="list-group-item"
  id="to-be-discussed"
>
  {item.value}
</div>
```

src/index.css
```diff
@@ -9,3 +9,13 @@ body {
   display: flex;
 
 }
+
+#to-be-counted{
+  list-style: decimal inside;
```
vit-svoboda: The counting doesn't seem to be working. All I get is 0.

src/components/List.jsx
```diff
@@ -1,18 +1,109 @@
...
+      ],
+      newItemText: '',
+    };
+  }
+
+  generateId = () => {
+    return ([1e7] + -1e3 + -4e3 + -8e3 + -1e11).replace(/[018]/g, c =>
+      (c ^ crypto.getRandomValues(new Uint8Array(1))[0] & 15 >> c / 4).toString(16)
+    );
+  };
```
vit-svoboda: Personally, I would get some npm package to do this sort of stuff for me, but I get it's in the instructions. This method, however, should not be placed in a component. Think of the single responsibility principle.

src/index.css
```diff
@@ -9,13 +9,11 @@ body {
   display: flex;
 
 }
-
-#to-be-counted{
-  list-style: decimal inside;
-  display: list-item;
-}
-
 .form-inline{
```
vit-svoboda: You should not hack Bootstrap classes like this. The right approach is either to build the elements with classes in a way the provided styles work properly or create all of the styles on your own. I believe you change the display to `inline-block` so the number is on the same line as the item editor/viewer. Maybe you could restructure the divs to something like this:
```
div /* the whole list */
  .list-group-item.form-inline
     /* the number */
     .input-group
        .form-control
        .btn.btn-primary
        .btn.btn-default
        .btn.btn-danger
  .list-group-item.form-inline
    /* ... */
```
On the other hand, the item viewer doesn't need to be a form-inline at all as it's not a form, nor does it Bootstrap make inline.

Btw. if the list and its items were `<ul>` and `<li>`, the markup would be a lot more semantic than just Alice in the world of divs. Nope, that doesn't work in English. Anyway, that's just food for thought.

src/components/defaultItemList.jsx
```diff
@@ -0,0 +1,25 @@
+const uuid4 = require('uuid/v4');
+
+export const defaultItemList = [
```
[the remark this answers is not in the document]
LindaLangerova: Yeah, I didn't realize, but when you say it, it makes perfect sense :-)

src/components/DisabledButton.jsx
```diff
@@ -0,0 +1,28 @@
...
+export const DisabledButton = (props) => {
+  DisabledButton.propTypes = {
+    buttonLabel: propTypes.string.required,
+    buttonType: propTypes.string.required,
+  };
+
+  return (
+    <span>
+      <a data-tip="React-tooltip">
+        <button
```
vit-svoboda: If this component is supposed to be a _button_, I'm not sure why it has to be hidden in a span and a hyperlink. From what I understand, the package for tooltips kind of dictates what the markup has to look like. That's not very nice. It's also one of the packages the owner tries to get rid of, see the readme in their GitHub repository. So in the real project, you probably wouldn't want to use it.
In this case, the HTML attribute [title](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title) could be all you need.
If you wanted something more fancy, you still need the control over the styles used for the tooltip (so they match the look & feel of the rest of your application), so you could go with a pure CSS solution - Kentico Cloud uses something like https://github.com/kazzkiq/balloon.css/
If you wanted to be super fancy, you could try http://tether.io/ but that's a huge overkill in my eyes.

No matter what you choose, keep in mind that tooltips are quite problematic on touch devices where there's no real mouse hover ;) The information in them should be also available in some other way.

src/components/DisabledButton.jsx
```diff
@@ -0,0 +1,28 @@
...
+          className={props.buttonType}
+          disabled
+        >
+          {props.buttonLabel}
+        </button>
+      </a>
+      <ReactTooltip place="top" type="info" effect="solid">
+        <span>
+          The text mustn't be empty
```
vit-svoboda: The "disabled button" should know nothing about texts and their validation rules.

src/components/List.jsx
```diff
@@ -23,19 +22,17 @@ export class List extends Component {
   };
```
vit-svoboda: I haven't noticed this the last time and you may need to change the `NewItemForm` back to a class component as a result - I'm sorry about that. The thing is, the list should not rely on the callback being called with an event argument from an input element. The `NewItemForm` should be able to create the updated text in any way it pleases now it has to construct the event arguments with it. It would be much better if `NewitemForm` called its prop `onTextChange` with just the target value, not the whole `e`.

src/components/EditedListItem.jsx
```diff
@@ -37,38 +40,44 @@ export class EditedListItem extends React.Component {
...
-          <div className="input-group-btn">
-            <button
+      <div className="input-group">
+        <input
+          className="form-control"
+          defaultValue={item.value}
+          onChange={this.onTextChanged}
+        />
+        <div className="input-group-btn">
+          {this.state.updatedValue === '' ?
```
vit-svoboda: Now you have the validation "logic" in two places - here and in the `NewItemForm`. I know there's not that much of the logic, but imagine you needed to add a check the value doesn't contain whitespaces only. You're likely to forget one of the occurrences.
You should create a util method for this.

src/components/DisabledButton.jsx
```diff
@@ -0,0 +1,28 @@
...
+  };
+
+  return (
+    <span>
+      <a data-tip="React-tooltip">
+        <button
+          className={props.buttonType}
+          disabled
+        >
+          {props.buttonLabel}
```
vit-svoboda: I'm not sure about the future of the `DisabledButton` component, knowledge won't hurt either way. Instead of passing a prop as an attribute, you can define child components, see the beginning of this [documentation](https://reactjs.org/docs/composition-vs-inheritance.html).
You limit the `buttonLabel` to a string in the `propTypes`, but the button could contain so much more.
```
<DisabledButton
  buttonType="Not really a button type, is it? ;)"
>
  <img src="http://dogecoin.com/imgs/dogecoin-300.png" />
</DisabledButton>
```

src/components/List.jsx
```diff
@@ -1,19 +1,72 @@
...
+        newItemText: '',
+      };
+    });
+  };
+
+  toggleEditing = (item) => {
+    this.setState((prevState) => {
+      const updatedList = prevState.listItems;
+      const index = updatedList.indexOf(item);
+      updatedList[index].isBeingEdited = !item.isBeingEdited;
```
vit-svoboda: This is the same case as was the `updateItemText`. The `item` and `updatedList[updatedList.indexOf(item)]` point to the same piece of memory and thus the shenanigans with indexes are absolutely pointless. You can just update the `item` or do it properly like you do in the `updateItemText` now.

src/components/List.jsx
```diff
@@ -1,19 +1,72 @@
...
+        isBeingEdited: false,
+      };
+      return {
+        listItems: [...prevState.listItems, newItem],
+        newItemText: '',
+      };
+    });
+  };
+
+  toggleEditing = (item) => {
```
vit-svoboda: This and the `deleteItem` method could accept only the `id` as a parameter. That's all the information they need to do their job.
We may discuss why exactly is it a bad thing to pass more information than they need to methods if you don't see the value in this change.

src/components/List.jsx
```diff
@@ -1,19 +1,72 @@
...
+  deleteItem = (item) => {
+    this.setState((prevState) => {
+      const listWithoutItem = prevState.listItems.filter(arrayItem => arrayItem !== item);
+      return {
+        listItems: listWithoutItem,
+      };
+    });
+  };
+
+  updateItemText = (item, newText) => {
```
vit-svoboda: This method as well doesn't need the whole `item` and just the `id` would do.

src/components/List.jsx
```diff
@@ -1,19 +1,75 @@
...
+            ? {
+              value: listItem.value,
+              id: itemId,
+              isBeingEdited: !listItem.isBeingEdited,
+            } : listItem)),
+      };
+    });
+  };
+
+  deleteItem = (item) => {
```
vit-svoboda: You missed this one when changing the parameters from the whole `item` to the much better `itemId`. You'll need to change the `filter` callback so it looks like the condition in the `map` in `toggleEditing` and `updateItemText`.

src/components/ListItem.jsx
```diff
@@ -0,0 +1,31 @@
...
+    onToggleEditing: PropTypes.func.isRequired,
+  };
+
+  toggleTextEditing = () => {
+    const { item, onToggleEditing } = this.props;
+    onToggleEditing(item.id);
+  };
+
+  render() {
+    const { item } = this.props;
```
vit-svoboda: You only use the `item.value` here and in the `EditedListItem` so you could go further with the destructuring.
```
const { value } = this.props.item;
```
This is not absolutely necessary, though ;)

src/components/List.jsx
```diff
@@ -1,19 +1,75 @@
...
+  addNewItem = (text) => {
+    this.setState((prevState) => {
+      const newItem = {
+        value: text,
+        id: generateId(),
+        isBeingEdited: false,
+      };
+      return {
+        listItems: [...prevState.listItems, newItem],
+        newItemText: '',
```
vit-svoboda: This state property doesn't seem to be used anymore. The `NewItemForm` has its own state for the text.

src/components/ListItem.jsx
```diff
@@ -0,0 +1,31 @@
...
+    return (
+      <div
+        className="form-group"
```
vit-svoboda: With the markup you have here, the `form-control-static` seems to a be semantically better class than the `form-group` and looks the same.

PR #81, opened 2017-11-06
Feature/task 2

kubmir:
Major changes - using immutable JS in the state - can be seen in commit with message KA-207 Storing of notes in immutable ordered map.


src/components/List.jsx
```diff
@@ -22,10 +23,9 @@ export class List extends PureComponent {
 
     this.setState((previousState) => {
       return {
-        notes: [
-          ...previousState.notes,
-          addNote,
-        ],
+        notes: previousState
+          .notes
+          .set(addNote.uid, addNote),
```
PetrSvirak: I would prefer `id` than `uid` - the later is too much of an acronym that suggests a type of the stored value while the former kinda industry standard replacement for "identifier".

kubmir:
Not using Immutable.Record was caused by a misunderstanding of the paragraph in Immutable js documentation. _Records ofters a persistently immutable alternative to plain JavaScript objects, however they're not required to be used within Immutable.js collections. In fact, the deep-access and deep-updating functions like getIn() and setIn() work with plain JavaScript Objects as well._ 

Now I understand the reason why to use Record within immutable.js ordered map.

src/models/NoteRecord.js
```diff
@@ -0,0 +1,7 @@
+import { Record } from 'immutable';
+
+export const NoteRecord = new Record({
```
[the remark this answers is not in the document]
PetrSvirak: Omit the suffix `Record` since it is a mere implementation detail (and also a name of the type - you would not write `textString` or `isEditActiveBool` just a few lines below).

PetrSvirak:
Next time you do a commit something like 9aa557da5b900c08dae2d957bb2277befb68c801: it is perfectly OK to split it into two, where one would state something like "Remove braces from simple lamdas" and the later "Clarify item model implementation". 

Also, while marked as done you seem to forget (or maybe lost?) code change based on this https://github.com/KenticoAcademy/kentico-onboarding-js/pull/81#discussion_r149716763

Despite these, we are all good :)

kubmir:
Thank you for the recommendation. 

Can we now merge this pull request or should I change a name from Note to something like ListItem ?

src/components/List.jsx
```diff
@@ -1,107 +1,93 @@
...
-  addNewNote = (newNoteText) => {
-    const noteToAdd = {
-      text: newNoteText,
-      uid: generateUid(),
-      isEditActive: false,
-    };
-
+  addNewNote = (newNoteText) =>
     this.setState((previousState) => {
+      const noteToAdd = Note({
```
PetrSvirak: You miss the `new` keyword here 😉 See [this article](http://raganwald.com/2014/07/09/javascript-constructor-problem.html) on JS ctor problem.

PetrSvirak:
Nope, since there is still the problem related to default record object from https://github.com/KenticoAcademy/kentico-onboarding-js/pull/81#discussion_r149716763. 
(and you miss one tiny `new` word there, but that's just a quick fix of a problem I just noticed, sorry for that)

PR #82, opened 2017-11-08
Feature/task 2

matus12:


src/utils/itemRecord.js
```diff
@@ -0,0 +1,7 @@
+import { Record } from 'immutable';
+
+export const ItemRecord = Record({
+  id: undefined,
```
PetrSvirak: This anonymous object of yours actually follows a [`Null object pattern`](https://en.wikipedia.org/wiki/Null_object_pattern). Please, read more about it and make the object follow the pattern (be correct but not carry any information).
There will quite a difference between `id` and `text` (from JS point of few, they are both `string`s, however, semantics is telling a completely different story)

src/components/PlainItem.jsx
```diff
@@ -1,14 +1,16 @@
...
 
 export class PlainItem extends PureComponent {
   static propTypes = {
     index: PropTypes.number.isRequired,
-    onClick: PropTypes.func.isRequired,
-    item: PropTypes.shape({
+    item: ImmutablePropTypes.contains({
       id: PropTypes.string.isRequired,
       text: PropTypes.string.isRequired,
+      isEdited: PropTypes.bool.isRequired,
```
PetrSvirak: Again, You shouldn't need this `propType` since `.isEdited` is never used in the component.

test/utils/generateItems.test.js
```diff
@@ -0,0 +1,43 @@
+import { generateItems } from '../../src/utils/generateItems';
+import { OrderedMap } from 'immutable';
+import { generateList } from '../../src/utils/initItemList';
+import * as Guid from 'guid';
+
+describe('generate items', () => {
```
PetrSvirak: These tests were not necessary, however, I really appreciate you wrote them anyway - that's the spirit! :)

src/components/Item.jsx
```diff
@@ -10,7 +11,7 @@ export class Item extends PureComponent {
     onCancel: PropTypes.func.isRequired,
     onTextClick: PropTypes.func.isRequired,
     index: PropTypes.number.isRequired,
-    item: PropTypes.shape({
+    item: ImmutablePropTypes.contains({
       id: PropTypes.string.isRequired,
       text: PropTypes.string.isRequired,
```
PetrSvirak: This prop is never used. I would personally remove, but since it is a bit higher than low-level (leaf) component, I would love to hear @vit-svoboda's opinion on this one...

src/components/List.jsx
```diff
@@ -1,77 +1,58 @@
...
-          ((item.id !== id)
-              ? item
-              : ({
-                id: item.id,
-                text: savedText,
-                isEdited: false,
-              })
-          ),
-        ),
+        .set(guid, new ItemRecord({
```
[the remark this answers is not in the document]
matus12: Not sure of correctness of my solution, I used setIn() which can set only one value at a time, so if I need to change e.g. 2, I call setIn 2x. Is that ok?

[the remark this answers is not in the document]
PetrSvirak: It would work, but wouldn't it be even better to use [`updateIn`](https://facebook.github.io/immutable-js/docs/#/updateIn) in combination with [`merge`](https://facebook.github.io/immutable-js/docs/#/merge)?

PR #85, opened 2017-11-13
Task1

RadoslavK:


RadoslavK:
Contains unused imports.

PetrSvirak:
It is OK to close PR if you change your mind and nice to do so if there is no other way of tracking the progress of your work.

However, You don't have to close PR when there is something missing. If you need to add more commits to it, simply push a commit to your feature branch on GitHub and it will appear in PR (if there is any). Commenting that something is still missing would be sufficient in this case. 

Moreover, since you track your progress in JIRA - once your JIRA task in Ready for Review step, we will attend to it, while it is In Progress, no review of your code will start - there was no need for closing the PR, it would be enough to move the task back In Progress state .

PR #86, opened 2017-11-13
Task1

RadoslavK:


src/components/ListItem.jsx
```diff
@@ -1,67 +1,67 @@
...
-            <button className="btn btn-danger" onClick={this.props.deleteClick}>Delete</button>
-          </div>
-        </li>
+        <ListItemEditMode
+          text={text}
+          number={number}
+          onSave={this.onSave}
+          onCancel={this.onCancel}
+          onDelete={this.onDelete}
+        />
```
vit-svoboda: Just return the component right away, no need to assign it to a variable just to return the variable couple lines later.

src/components/ListItemEditMode.jsx
```diff
@@ -0,0 +1,69 @@
+import React, { Component } from 'react';
+import { PropTypes } from 'prop-types';
+
+export class ListItemEditMode extends Component {
```
vit-svoboda: `EditMode` feels more like an `enum` value than a component name. An `Editor` might be better, but I don't think it's really necessary to change this.

src/components/ListItemEditMode.jsx
```diff
@@ -0,0 +1,69 @@
...
+          type="text" value={text}
+          onChange={this.onInputChange}
+        />
+
+        <button
+          className="btn btn-primary"
+          onClick={this.onSave}
+        >
+          Save
+        </button>
```
vit-svoboda: I'm still able to click this button even if the validation of the current value failed. The click doesn't do anything, but as I said earlier, that's not a nice UX.

src/components/ListItem.jsx
```diff
@@ -0,0 +1,77 @@
+import React, { Component } from 'react';
+import { PropTypes } from 'prop-types';
+import { ListItemEditMode } from './ListItemEditMode';
+import { inputIsNotEmpty } from '../utils/validation.js';
+
+export class ListItem extends Component {
```
vit-svoboda: You should always prefer `PureComponent` to `Component`, it optimizes the rendering so the component is not redrawn when the props didn't change. Or something like that I believe, check the documentation to be sure.

src/components/ListItem.jsx
```diff
@@ -0,0 +1,77 @@
...
+    return listItemContent;
+  }
+}
+
+ListItem.propTypes = {
+  item: PropTypes.object.isRequired,
+  number: PropTypes.number.isRequired,
+  onSave: PropTypes.func.isRequired,
+  onDelete: PropTypes.func.isRequired,
+};
```
vit-svoboda: You should not export the component before you're done with it (i.e. the `propTypes` haven't been set yet). Here's the place for `export { ListItem };`

src/components/ListItem.jsx
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  onSave = (text) => {
+    const { id } = this.props.item;
+
+    if (inputIsNotEmpty(text)) {
+      this.setState({ isBeingEdited: false });
+      this.props.onSave(id, text);
+    }
+  }
```
vit-svoboda: This is an assignment and as such should be terminated with a semicolon. I'm sorry if my arrow function example missed this.

src/components/NewItemForm.jsx
```diff
@@ -0,0 +1,47 @@
...
+        <input
+          className="form-control"
+          type="text"
+          placeholder="Item name cannot be empty"
+          value={this.state.newItemText}
+          onChange={this.onInputChange}
+        />
+
+        <button
+          className="btn btn-secondary"
```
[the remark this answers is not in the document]
RadoslavK: It was used just to set the button color but yes it makes more sense to make it primary to avoid confusion:+1:

RadoslavK:
Commit history changed due to adding missing JIRA issue code. The last review commit was [KA-220 Refactor requested changes (36272c2)](https://github.com/KenticoAcademy/kentico-onboarding-js/pull/86/commits/36272c207c67ccdc2c78f8b73d66793f2fde6bd6)

src/components/ListItemEditMode.jsx
```diff
@@ -0,0 +1,100 @@
...
+        </label>
+
+        <input
+          className="form-control col-md-5"
+          type="text"
+          value={text}
+          placeholder={NEW_ITEM_TEXT_PLACEHOLDER}
+          onChange={this.onInputChange}
+          onKeyPress={this.onKeyPress}
+          ref="input"
```
vit-svoboda: Up until quite recently, the string refs were discouraged, yet not obsolete. As it seems, the [new documentation](https://reactjs.org/docs/refs-and-the-dom.html#legacy-api-string-refs) went a step further.

src/components/ListItemEditMode.jsx
```diff
@@ -0,0 +1,100 @@
...
+    this.state = {
+      text,
+    };
+  }
+
+  putFocusAtTheEndOfText = () => {
+    const { input } = this.refs;
+    const { value } = input;
+    input.value = '';
+    input.value = value;
```
vit-svoboda: While this likely works in the most of the browsers, using [setSelectionRange](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/setSelectionRange) would say better what you're doing.

src/components/ListItemEditMode.jsx
```diff
@@ -0,0 +1,100 @@
...
+
+  onInputChange = (e) => this.setState({ text: e.target.value });
+
+  onSave = () => {
+    const { text } = this.state;
+    this.props.onSave(text);
+  };
+
+  onKeyPress = (target) => {
+    if (target.charCode === 13) {
```
vit-svoboda: When dealing with keyboard input, you should use the `key` property instead, see the [documentation](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/charCode).

src/components/ListItemEditMode.jsx
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  onInputChange = (e) => this.setState({ text: e.target.value });
+
+  onSave = () => {
+    const { text } = this.state;
+    this.props.onSave(text);
+  };
+
+  onKeyPress = (target) => {
```
vit-svoboda: The keyboard handling is quite duplicate and there are many packages available for it on npm. You could try using e. g. [react-hotkeys](https://www.npmjs.com/package/react-hotkeys).

src/components/ListItemEditMode.jsx
```diff
@@ -0,0 +1,100 @@
...
+
+  render() {
+    const { text } = this.state;
+    const { number } = this.props;
+    const enableSaveButton = !textIsEmpty(text);
+
+    return (
+      <div className="form-inline">
+        <label className="col-form-label">
+          {number}{'. '}
```
vit-svoboda: You can do normal operations in the JSX `{}` block, e.g. `{number + '. '}`, but what might seem even cooler to you is the string interpolation available in javascript: ``{`${number}. `}`` - admittedly, the backtick vs. single quote is not as visually different as it could be, though.

src/utils/validation.js
```diff
@@ -0,0 +1,5 @@
+export function textIsEmpty(input) {
+  return input === '';
+}
+
+export const NEW_ITEM_TEXT_PLACEHOLDER = 'Item name cannot be empty';
```
vit-svoboda: The `NEW_ITEM_TEXT_PLACEHOLDER` has hardly anything to do with `validation`. Also, you don't use it only as a placeholder for the new item, do you?

src/components/NewItemForm.jsx
```diff
@@ -40,7 +40,7 @@ class NewItemForm extends PureComponent {
         <input
           className="form-control col-md-5"
           type="text"
-          placeholder={NEW_ITEM_TEXT_PLACEHOLDER}
+          placeholder="Item name cannot be empty"
```
vit-svoboda: Yes, removing the constant is a reasonable solution - the texts happen to be the same right now, but represent something different.

src/components/ListItemEditMode.jsx
```diff
@@ -0,0 +1,110 @@
...
+          handlers={handlers}
+        >
+          <input
+            className="form-control WideInput"
+            type="text"
+            value={text}
+            placeholder="Item name cannot be empty"
+            onChange={this.onInputChange}
+            autoFocus={true}
+            onFocus={this.putFocusAtTheEndOfText}
```
vit-svoboda: Although this solution is very elegant and I genuinely like it, it has one usability flaw.
- Click on an item, making it editable.
- Click somewhere else so it loses focus.
- Click back to the textbox, to some specific place in the text.

One might expect at this point the cursor to be placed where I clicked on the text. But the cursor is at the end of the text instead. Do you see the problem?
The `componentDidMount` was called only when the item editing form appeared. This `onFocus` is called every time the textbox gains focus.

src/css/WideInput.css
```diff
@@ -0,0 +1,5 @@
+.WideInput {
```
vit-svoboda: First off, all of the CSS classes you use are named using the kebab-case. Except for this one that uses PascalCase. You see the problem, right?
Anyway, we discussed that you should stick to the Bootstrap classes since you chose to use it. (It may have been with someone else and if so, I'm sorry about this, but you still should.) Seems you didn't need an extra class before you started using the `HotKeys` component. Did it break your markup? Can you fix it without messing with the styles? It renders an unstyled div if I remember correctly. Maybe you'll need to push it out of the "form-inline" div for the styles to work as they used to.

src/components/ListItem.jsx
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    return (
+      <div onMouseUp={this.onItemClick}>
+        <span>
+          {number + '. '}
+        </span>
+        <span>
+          {text}
+        </span>
```
vit-svoboda: Do you want the number on a new line on super narrow screens? Or what is the purpose of these spans?

src/components/ListItem.jsx
```diff
@@ -0,0 +1,80 @@
...
+      </div>
+    );
+  }
+}
+
+ListItem.propTypes = {
```
vit-svoboda: This delayed assignment of `propTypes` is necessary for stateless functional components, but when you have a class, you can place it inside as a static property. Usually at the beginning, above the constructor.

src/components/ListItemEditMode.jsx
```diff
@@ -0,0 +1,113 @@
...
+        </div>
+      </HotKeys>
+    );
+  }
+}
+
+ListItemEditMode.propTypes = {
+  number: PropTypes.number.isRequired,
+  startOffset: PropTypes.number.isRequired,
+  endOffset: PropTypes.number.isRequired,
```
vit-svoboda: What offset? I know it's the selection range offset, but in this context, it's not clear. You should add more information to the prop names - something like `selectionRangeStartOffset` or just `selectionRangeStart` if it's not misleading.

src/components/List.jsx
```diff
@@ -1,35 +1,73 @@
...
-        {/* TODO: You can delete the assignment part once you do not need it */}
+    const listItems = this.state.items.map((item, index) =>
+      <li
+        className="list-group-item"
+        key={item.id}
+      >
         <div className="row">
-          <div className="col-sm-12 text-center">
-            <TsComponent name="𝕱𝖆𝖓𝖈𝖞" />
+          <div className="col">
```
vit-svoboda: These 2 divs seem pretty useless to me. Same goes for the `NewItemForm` row.

PR #89, opened 2017-11-14
Feature/task 3

kubmir:


test/reducers/notes/list/addListMember.test.js
```diff
@@ -0,0 +1,46 @@
+import { isAddingNote } from '../../../../src/reducers/notes/list/isAddingNote';
```
vit-svoboda: The test files should be named after the thing they test, in this case, `isAddingNote.test.js`

test/reducers/notes/list/addListMember.test.js
```diff
@@ -0,0 +1,46 @@
...
+import {
+  prepareAddListMemberInitialState,
+} from '../../../testUtils/prepareTestData';
```
vit-svoboda: I'd declare this in the test, there's no reuse value. Zuzka even advocates that the initial state of a reducer should be declared only in the default value of the state parameter of the reducer.

test/reducers/notes/list/addListMember.test.js
```diff
@@ -0,0 +1,46 @@
...
+  let initialState;
+
+  beforeEach(() => {
+    initialState = prepareAddListMemberInitialState();
+  });
+
+  it('should return previous state if unknown action is dispatched', () => {
+    const unknownAction = {
+      type: 'Test action',
+      payload: {},
```
vit-svoboda: The action doesn't have to carry a payload. So this line is unnecessary.

test/selectors/notes/list/listOfNotes.test.js
```diff
@@ -0,0 +1,41 @@
...
+    expect(actualNote).toEqual(expectedNote);
+  });
+
+  it('getAllIds test', () => {
+    const notes = prepareNotesInitialState();
+    const expectedIds = [1, 2];
+
+    const actualIds = getAllIds(notes);
+
+    expect(actualIds).toEqual(expectedIds);
```
vit-svoboda: While this test is fine, you should also test the memoization - that the function returns the same object if you call it multiple times.

src/components/List.jsx
```diff
@@ -1,120 +1,31 @@
...
-              <AddListMember
-                onAddClick={this.addNewNote}
-                onInputFocus={this.onIsAddListMemberFocus}
-                onInputBlur={this.onIsAddListMemberBlur}
-                isInputFocused={this.state.isAddListMemberFocused}
-              />
-            </li>
-          </ul>
-        </div>
+      <div>
```
vit-svoboda: You render this `List` in a `<ul>` element, that should only contain `<li>` elements - [don't take my word for it](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul). If you threw your output markup at a validator, I believe you'd get an error.

The solution here is either upgrading React to version 16 so you can [return fragments from a render](https://reactjs.org/blog/2017/09/26/react-v16.0.html#new-render-return-types-fragments-and-strings). Or you can restructure your components so the whole `<ul>` with all `<li>` are together in a single component.

kubmir:
Now, I store an id of note to the database as noteId, Note object contains id property and other ids in the application are named noteId. 

I agree with possible renaming of ListMembers to Notes. I will do it in task 4 if it is possible because now it would be harder to rename files in task 3 and also in task 4 branch - I would do the renaming twice.

PR #94, opened 2017-11-28
Feature/task 4

kubmir:


src/reducers/notes/list/listOfNotes.ts
```diff
@@ -37,17 +22,33 @@ const setNoteToState = (state, action) => {
...
-  return state.update(noteId, note => note.merge({
+  return state.update(noteId, note => note.with({
     isEditActive: newEditingMode,
   }));
 };
+
+export const listOfNotes = (state = OrderedMap<string, Note>(), action: IAction): OrderedMap<string, Note> => {
+  switch (action.type) {
+    case ADD_NOTE:
+    case UPDATE_NOTE:
```
Neytus: `UPDATE_NOTE` should have its own method, similar to `changeEditingMode`.

src/reducers/notes/list/listOfNotes.ts
```diff
@@ -0,0 +1,65 @@
...
+} from '../../../constants/actionTypes';
+import { IAction } from '../../../models/IAction';
+import { Guid } from '../../../@types/globals';
+
+const setNoteToState = (state: OrderedMap<string, Note>, payload: { noteId: Guid, text: string }): OrderedMap<string, Note> => {
+  const { noteId, text } = payload;
+  const noteToAdd = new Note({
+    id: noteId,
+    text,
+    isEditActive: false,
```
Neytus: As your `Note` model constructor requires `Partial<INote>` parameters, you can now omit `isEditActive` attribute (it is false by default).

src/components/NonEmptyInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+  _onCancelFocusOfInput = (): void => {
+    if (this.props.onCancelEditing) {
+      this.props.onCancelEditing();
+    } else {
+      this.textInput.blur();
+    }
+  };
+
+  _onSaveChanges = (): void => {
+    if (this.props.text) {
```
Neytus: This `if` should be checking a different property - you want to disable the functionality depending on whether the input is valid or not. You can utilize your `isError` prop.

src/utils/isNoteValid.ts
```diff
@@ -0,0 +1,2 @@
+export const isNoteValid = (currentNoteText: string): boolean =>
+  !!currentNoteText && currentNoteText.length > 0;
```
Neytus: Is there a reason why you didn't use the `trim` function in the validation? I am still able to add a visually blank note.

PR #95, opened 2017-11-28
KA-214 Feature/task 3

LindaLangerova:


src/components/List.jsx
```diff
@@ -1,102 +1,31 @@
...
+        itemsMap.valueSeq().map((item, index) => (
+          <div
+            className="list-group-item form-inline"
+            key={item.id}
+          >{index + 1}
+            {'. '}
+            {item.isBeingEdited ?
+              <EditedListItem
+                text={item.text}
+                onSave={(newText) => updateItemText(item.id, newText)}
```
vit-svoboda: More lambdas in the `render` 😠 The `EditedListItem` and `ListItem` will need to get the `item.id` in their props and then pass it to the action handlers they get. You could make a container for this so you don't need to convert `ListItem` to a component class (as you'll need a class method there). The `mapDispatchToProps` of the container could inject the `id` from own props to the dispatched action. If you feel you don't understand all of what I wrote here, we can discuss it in person. It'll be more efficient than trying to describe everything here perfectly.

src/components/ListItem.jsx
```diff
@@ -1,31 +1,17 @@
-import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
+import React from 'react';
 
-export class ListItem extends PureComponent {
+export const ListItem = ({ onClick, text }) => (
```
vit-svoboda: You should `export` the component _after_ you're done with it - that includes the `propTypes`. The change someone would use the component before you set the prop type validation up is close to zero, but why risk it?

src/constants/defaultListItems.js
```diff
@@ -0,0 +1,11 @@
+import { addItem } from '../actions/actionCreators';
+import { generateId } from '../utils/generateId';
+
+const defaultListItems = (store) => {
+  store.dispatch(addItem(generateId(), 'Make a coffee'));
+  store.dispatch(addItem(generateId(), 'Master React'));
+  store.dispatch(addItem(generateId(), 'Learn Redux'));
+  store.dispatch(addItem(generateId(), 'Help making Draft awesome'));
```
vit-svoboda: Dispatching changes to modify the initial state is not necessary. You can pass the initial state as a second argument to the [createStore](https://redux.js.org/docs/api/createStore.html) function. In that case, the middleware would be the third argument.

src/reducers/items/newItemText.js
```diff
@@ -0,0 +1,18 @@
+import {
+  ADD_ITEM,
+  UPDATE_NEW_ITEM_TEXT,
+} from '../../constants/actionTypes';
+
+const DEFAULT_VALUE = '';
+
+export const newItemText = (state = DEFAULT_VALUE, action) => {
```
vit-svoboda: You miss tests for this reducer. It's rather primitive, but so can be the tests.

src/utils/checkEmptiness.js
```diff
@@ -0,0 +1,2 @@
+
+export const checkEmptiness = (text) => !text;
```
vit-svoboda: Is `'   '` valid value for you? If not, [trim](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim) can help you.

vit-svoboda: Also, didn't we discuss the `checkEmptiness` being a rather confusing name?

src/containers/AddNewItemContainer.js
```diff
@@ -0,0 +1,18 @@
...
+const mapStateToProps = ({ newItemText }) => ({
+  newItemText,
+});
+
+const mapDispatchToProps = (dispatch) => ({
+  onAdd: (value) => dispatch(addItem(generateId(), value)),
+  onNewTextChange: (newText) => dispatch(updateNewItemText(newText)),
+});
+
+export const AddNewItem = connect(mapStateToProps, mapDispatchToProps)(AddNewItemComponent);
```
vit-svoboda: The file is still called `AddNewItemContainer` even though it exports `AddNewItem` without the "Container" suffix.

src/reducers/items/byId/byId.js
```diff
@@ -0,0 +1,38 @@
+import {
+  ADD_ITEM,
+  UPDATE_ITEM_TEXT,
+  DELETE_ITEM,
+  TOGGLE_EDITING,
+} from '../../../constants/actionTypes';
+import { OrderedMap } from 'immutable';
+import { Item } from '../../../models/Item';
+
+export const byId = (state = new OrderedMap(), action) => {
```
vit-svoboda: I don't think the `byId` reducer needs to sit in a `byId` folder just to group it with its tests. After all, maybe you could merge it with the `items` reducer to create something like `itemsById` property in the store. If you don't plan to add more properties to the `items` sub-state of course.

src/containers/ListContainer.js
```diff
@@ -0,0 +1,22 @@
...
+  deleteItem,
+  toggleEditing,
+  updateItemText,
+} from '../actions/actionCreators';
+import { connect } from 'react-redux';
+import { List as ListComponent } from '../components/List';
+
+function mapStateToProps(state) {
+  return {
+    itemsMap: state.items.byId,
```
vit-svoboda: This should be the place where we introduce you to the concept of memoization. Remind me to come up with some complication that will eventually require you to memoize something. #LazyReviewerWantsToGoHome

test/App.test.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import { App } from '../src/App.jsx';
 
-it('renders without crashing', () => {
+xit('renders without crashing', () => {
```
vit-svoboda: Fix the test or delete it. I think you need to wrap the `App` in a redux store `Provider` what should be fairly easy.

src/components/EditedListItem.jsx
```diff
@@ -61,19 +44,19 @@ export class EditedListItem extends PureComponent {
             data-balloon-pos="up"
             className="btn btn-primary"
             disabled={isEmpty}
-            onClick={this.saveNewText}
+            onClick={() => onSave(this.state.updatedValue)}
```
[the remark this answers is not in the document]
PetrSvirak: I would dare to suggest the original `saveNewText` name or something similar. Since it is not a callback but just a private method, naming should IMO be quite common and descriptive, as a bonus, it reads better: `onClick={this.saveNewText}` → "on click, save new text" 😇

src/reducers/items/byId/byId.js
PetrSvirak: I believe that `items` here is the name of the "app" (in Draft terminology) so it is not a good idea to merge it with `byId` (considering both structure and future tasks)

src/containers/ListItem.js
```diff
@@ -0,0 +1,9 @@
+import { connect } from 'react-redux';
+import { ListItem as ListItemComponent } from '../components/ListItem';
+
+const mapStateToProps = (state, { itemId }) => ({
+  id: itemId,
+  isBeingEdited: state.items.byId.get(itemId).isBeingEdited,
+});
+
+export const ListItem = connect(mapStateToProps, null)(ListItemComponent);
```
vit-svoboda: The `null` is unnecessary, `connect` can handle having just 1 argument.

vit-svoboda: Also, it's conventional to place containers to files with `jsx` extension (or `tsx` later on). Most likely for consistency, as the usages of a component or a container are exactly the same and containers deal with JSX markup transitively. On the other hand, there's no need for the `Item` record to be placed in a `jsx` file. I haven't noticed the extensions earlier, my bad.

src/utils/isNotText.js
```diff
@@ -0,0 +1,2 @@
+
+export const isNotText = (text) => (!text || text.trim() === '');
```
vit-svoboda: This name is a bit harsh. The argument is text - perhaps not a valid for some purposes, but a text nonetheless :P When designing functions, you usually want to avoid the negation in the name, but that's more of a general suggestion than an instruction for you to rename the function.
Also, the comparison with an empty string is a bit unnecessary since the empty string is already falsy in JavaScript. So you can write just `!text || !text.trim()`, though I understand that these JavaScript internals can make the code quite obscure if you overuse them.

src/constants/defaultListItems.js
```diff
@@ -0,0 +1,48 @@
...
+    }),
+  ],
+  [
+    id4,
+    new Item({
+      id: id4,
+      text: 'Help making Draft awesome',
+    }),
+  ],
+];
```
vit-svoboda: You could rewrite this to be substantially shorter:
```
const defaultListItems = [
  'Make a sandwich',
  'Master Redux',
  'Learn TypeScript',
  'Help making Draft good enough',
].map(text => {
  const id = generateId();
  return [ id, new Item({ id, text })];
});
```
How cool is that!

src/components/EditedListItem.jsx
```diff
@@ -1,58 +1,35 @@
...
-  deleteItem = () => {
-    const { item, onItemDeletion } = this.props;
-    onItemDeletion(item.id);
+  static propTypes = {
+    itemText: PropTypes.string.isRequired,
+    textUpdate: PropTypes.string.isRequired,
+    onSave: PropTypes.func.isRequired,
+    onDelete: PropTypes.func.isRequired,
+    onCancel: PropTypes.func.isRequired,
+    textUpdateChange: PropTypes.func.isRequired,
```
vit-svoboda: Maybe I'd consider renaming this callback as well. Especially since there's `onNewTextChange` in the component for adding items.

src/containers/List.jsx
```diff
@@ -0,0 +1,12 @@
+import { connect } from 'react-redux';
+import { List as ListComponent } from '../components/List';
+import { memoized } from '../selectors/selector';
+
+const mapStateToProps = (state) => {
+  return {
+    ids: memoized(state.items.byId.keySeq()),
```
vit-svoboda: I should have been clearer about the selector - it's supposed to be a projection from the redux state (or its part) to the props. The selector should accept the whole `state` or at least the `byId` map (what may actually be better) and memoize the selection of keySeq(). Now, when it'll actually do something, you'll find a proper name more easily for it. I could imagine something like `selectItemIds` or `selectitemIdsMemoized` if you really wanted to point out the memoization part.

src/selectors/selector.js
```diff
@@ -0,0 +1,3 @@
+import memoize from 'fast-memoize';
+
+export const memoized = memoize((ids) => ids);
```
vit-svoboda: Why would you export `memoized` from a file called `selector`? ;)

src/selectors/selector.test.js
```diff
@@ -0,0 +1,10 @@
+import { memoized } from './selector';
+
+describe('memoized', () => {
+  it('return the same object when is given two different objects with the same value', () => {
+    const firstArr = new Array([1, 2, 3]);
+    const secondArr = new Array([1, 2, 3]);
+
+    expect(memoized(firstArr)).toBe(memoized(secondArr));
```
vit-svoboda: Good you replaced the `toEqual` with `toBe` ;)

PR #96, opened 2017-11-30
Feature/task 5

milan009:


onboarding-task/src/utils/fetchJsonResponse.ts
```diff
@@ -0,0 +1,15 @@
+export interface IFetchDependencies {
+  fetch: (input: RequestInfo, init?: RequestInit) => Promise<Response>;
+  input: RequestInfo;
+  init?: RequestInit;
+}
+
+export const fetchJsonResponse = async <ReturnType>(dependencies: IFetchDependencies): Promise<ReturnType> => {
```
vit-svoboda: This file looks all broken in my WebStorm. Maybe related to the `ResponseInfo` problem I mentioned elsewhere, but it really doesn't like the `async` there either.

Anyway, these `dependencies` are not really dependencies for some other call as you don't return a method. Maybe you should treat this utils function as a first-class thunk, get it its dependencies and then inject it to the rest of the thunks.

onboarding-task/src/reducers/list/itemFlagsMapReducer.ts
```diff
@@ -4,30 +4,57 @@ import { ItemFlags } from '../../models/ItemFlags';
...
     case ITEM_MAKE_EDITABLE:
     case ITEM_CHANGE_CANCELLED:
-    case ITEM_CHANGE_SAVED: {
+    case DELETE_REQUEST_STARTED:
+    case UPDATE_REQUEST_STARTED:
+    case UPDATE_REQUEST_SUCCESS:
+    case UPDATE_REQUEST_FAIL:
+    case DELETE_REQUEST_FAIL:
+    case CREATE_REQUEST_FAIL: {
       const flagsToEdit = state.get(action.payload.id);
```
vit-svoboda: Here you could leverage the Map's [update](https://facebook.github.io/immutable-js/docs/#/Map/update) method:
```
return state.update(action.payload.id, flagsInfo =>
  itemFlagsReducer(flagsInfo, action));
```

onboarding-task/src/reducers/list/itemFlagsMapReducer.ts
```diff
@@ -4,30 +4,57 @@ import { ItemFlags } from '../../models/ItemFlags';
...
+    case CREATE_REQUEST_SUCCESS: {
+      const newItem = new ItemFlags({isStored: true});
+      state = state.remove(action.payload.formerId);
+
+      return state.set(action.payload.item.id, newItem);
+    }
+
+    case FETCH_REQUEST_SUCCESS: {
+      action.payload.items.map((item: IItemDTO) => {
+        state = state.set(item.id, new ItemFlags({isStored: true}));
```
vit-svoboda: Changing the parameter of a function you're in is considered rude. Reduce the `payload.items` to a new map or use a new variable.

onboarding-task/src/reducers/list/itemFlagsMapReducer.ts
```diff
@@ -4,30 +4,57 @@ import { ItemFlags } from '../../models/ItemFlags';
...
+    case CREATE_REQUEST_STARTED: {
+      const newItem = new ItemFlags({isStored: false});
+      return state.set(action.payload.optimisticId, newItem);
+    }
+
+    case CREATE_REQUEST_SUCCESS: {
+      const newItem = new ItemFlags({isStored: true});
+      state = state.remove(action.payload.formerId);
+
+      return state.set(action.payload.item.id, newItem);
```
vit-svoboda: You can fluently chain these calls:
```
return state
  .remove(action.payload.formerId)
  .set(action.payload.item.id, newItem);
```

onboarding-task/src/actions/thunkActionsCreators.ts
```diff
@@ -0,0 +1,30 @@
...
+import * as fetch from 'isomorphic-fetch';
+
+import { ThunkAction } from '../interfaces/IAction';
+import { ItemData } from '../models/ItemData';
+import { deleteItemThunkFactory } from './thunkFactories/deleteItemThunkFactory';
+import { fetchItemsThunkFactory } from './thunkFactories/fetchItemsThunkFactory';
+import { postItemThunkFactory } from './thunkFactories/createItemThunkFactory';
+import { updateItemThunkFactory } from './thunkFactories/updateItemThunkFactory';
+
+export const deleteItemThunk: (id: string) => ThunkAction = deleteItemThunkFactory({
```
vit-svoboda: This is `deleteItem`, nobody gives a fsck it's a thunk. Yes, it may take a while, but the middleware will take care of that. Good names otherwise, I'm glad you abstracted the HTTP protocol.

onboarding-task/src/components/RetryItem.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+export interface IRetryItemDataProps {
+  item: IViewItem;
+}
+
+export interface IRetryItemCallbackProps {
+  onRetry: (action: ThunkAction) => void;
+}
+
+export const RetryItem: React.StatelessComponent<IRetryItemDataProps & IRetryItemCallbackProps> = (props) => (
```
vit-svoboda: Maybe something like `ItemWithRetry` or `ItemWithFailedOperationPending` would be a better name.

onboarding-task/src/components/AddItem.tsx
```diff
@@ -29,7 +29,7 @@ class AddItem extends React.PureComponent<IAddItemCallbackProps, IAddItemState>
   }
 
   _onChange = (event: FormEvent<HTMLInputElement>): void => {
-    const setStateText = (currentText: string) => (() => ({ currentText }));
+    const setStateText = (currentText: string) => (() => ({currentText}));
     this.setState(setStateText(event.currentTarget.value));
```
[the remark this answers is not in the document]
milan009: Back at task-3 in my first pull request, I got this piece of advice from Zuzka, and I have attempted to adhere to it sever since.
image
Although, I have re-read the [article about using the functional setState](https://medium.freecodecamp.org/functional-setstate-is-the-future-of-react-374f30401b6b), and I must admit I did not really find out what is the benefit of using it over regular setState if not working with previous state. 

Are there any drawbacks that should distract from using it everywhere though?

onboarding-task/src/reducers/list/itemFlagsMapReducer.ts
milan009: Good point, thanks!

onboarding-task/src/reducers/list/itemFlagsMapReducer.ts
milan009: If I am not mistaken, JS map() does not change the array it is called upon, but instead returns a new array containing elements that have had the function applied on, right? I should not be changing anything in this method - I only want to modify the state for each item present in the payload. 

Truly though, using map for this purpose was probably not the wisest choice. I have swapped it with .forEach() for time being.

onboarding-task/src/reducers/status/fetchingReducer.ts
```diff
@@ -0,0 +1,20 @@
+import { IAction } from '../../interfaces/IAction';
+import { FETCH_REQUEST_STARTED } from '../../actions/actionTypes';
+import { FETCH_REQUEST_SUCCESS } from '../../actions/actionTypes';
+import { FETCH_REQUEST_FAIL } from '../../actions/actionTypes';
+
+const defaultState = false;
+
+export const fetchingReducer = (state: boolean = defaultState, action: IAction): boolean => {
```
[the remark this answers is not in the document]
milan009: I agree, though I have to say that it seems more clear to me this way - as I am calling the reducer for single item within the reducer for whole map, having it explicitly say it is a reducer, not just some unknown "item" function feels better.

Would renaming them at import be more suitable?

onboarding-task/src/utils/parsing.ts
```diff
@@ -0,0 +1,14 @@
...
+
+import { ItemsDataMap } from '../reducers/list/itemsReducer';
+import { ItemData } from '../models/ItemData';
+import { IItemDTO } from '../interfaces/IItemDTO';
+
+export const parseAPIResponseJson = (json: IItemDTO[]): ItemsDataMap => {
+    let parsedItems = OrderedMap<string, ItemData>();
+    json.map((item: IItemDTO) => {
+      parsedItems = parsedItems.set(item.id, new ItemData(item));
+    });
```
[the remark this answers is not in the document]
milan009: That indeed is lame. Boosted the code sexiness to 150% by using reduce.

onboarding-task/package.json
```diff
@@ -12,6 +14,7 @@
     "@types/react-dom": "^0.14.20",
     "@types/react-redux": "^4.4.47",
     "@types/redux": "^3.6.0",
+    "@types/redux-thunk": "^2.1.0",
```
vit-svoboda: The `redux-thunk` package contains own typings since version 2.2.0 that you're using. So this `@types` package only contains readme saying you should not use it anymore.

onboarding-task/src/reducers/list/itemFlagsReducer.ts
```diff
@@ -14,9 +19,24 @@ export const itemFlagsReducer = (state: ItemFlags = defaultState, action: IActio
...
+    case DELETE_REQUEST_STARTED:
+    case UPDATE_REQUEST_STARTED:
+      return state.typedMerge({isBeingEdited: false, isStored: false, requestError: null});
+
+    case UPDATE_REQUEST_SUCCESS:
+      return state.typedMerge({isStored: true, isBeingEdited: false});
+
+    case UPDATE_REQUEST_FAIL:
+    case DELETE_REQUEST_FAIL: {
+      return state.typedMerge(new ItemFlags({isStored: true, requestError: action.payload}));
```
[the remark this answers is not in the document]
milan009: Without instantiating `ItemFlags`, I was unable to get my test to pass, as the process of merging state always converted the `requestError` object passed from the payload into an Immutable structure - something that I had not anticipated beforehand and that failed the `toEqual()` assertion.

After digging around this a little, I found out that this *was* intended behavior, as the documentation within Immutable.d.ts file states that: 

"*(...) If any of the values provided to `merge` are not Iterable (would return false for `Immutable.Iterable.isIterable`) then they are deeply converted via `Immutable.fromJS` before being merged. However, if the value is an Iterable but includes non-iterable JS objects or arrays, those nested values will be preserved.*"

This way, when an `ItemFlags` (which is an Immutable Record) object is passed to the merge function, the `requestError` remains a plain JS object. However, if I pass in a complex JS object, it is converted into an Immutable structure, thus failing the tests. 

It seems that this behavior was questioned in [this issue](https://github.com/facebook/immutable-js/issues/1293) and changed in [this PR](https://github.com/facebook/immutable-js/pull/1339) to master, yet it does not seem to be present in the package distributed via NPM so far.

onboarding-task/src/reducers/list/itemFlagsMapReducer.ts
milan009: While using the `update` method would make things clearer, it seems to create a new record in the map if one with given ID does not exist already. I don't think that this reducer should work that way.

onboarding-task/src/reducers/list/itemFlagsReducer.ts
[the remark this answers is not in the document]
vit-svoboda: That is an interesting problem. There hasn't been any release of immutable.js in quite some time, hence the missing changes in the package. All PRs accepted in the last year or so will be only in the version 4.0 that's still in beta.
However, for an object to be really immutable, all of its properties need to be immutable as well. Transitively. Otherwise, we'd be able to do something like this:
```
const immutableRecord = new ItemFlags({
  requestError = new Error('The original error'),
});
immutableRecord.requestError.message = 'Undesired mutation right here';
```
What pretty much defeats the whole purpose of the immutability exercise.

So in retrospect, perhaps storing the whole `Error` object may be a bad idea and you should either just extract the message from it or intentionally convert it to an immutable record of your own.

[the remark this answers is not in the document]
milan009: A fair point indeed, I'll have a look at it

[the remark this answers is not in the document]
milan009: I have introduced a new `RequestError` Immutable record which should do the trick now.

onboarding-task/src/reducers/list/itemFlagsMapReducer.ts
vit-svoboda: Apparently, the `update` callback is called with `undefined` if an item with given `id` doesn't exist in the map. And what does your `itemFlagsReducer` do if it gets an undefined previous state? (See the default value of the parameter)

An easy fix is to add _null_ check to the callback. Like this:
```
return state.update(action.payload.id, flagsInfo =>
  flagsInfo && itemFlagsReducer(flagsInfo, action));
```

milan009: This is brilliant, thanks! Had to look up how does `&&` behave on non-boolean convertible objects as it seemed to be doing just some weird magic, but it now works perfectly. Thanks 👍

milan009:
Whoops, last 6 commits are titled with incorrect task number - they all refer to KA-197.

PR #97, opened 2017-12-01
Feature/task 4

matus12:


src/components/todo-list/Input.tsx
```diff
@@ -1,39 +1,51 @@
...
+  constructor(props: IProps) {
     super(props);
 
     this.state = ({
       isFocused: false,
     });
   }
 
-  focus = () => {
+  focus = (): void => {
```
vit-svoboda: With standard HTML components, there's usually a `focus` method and an `onFocus` handler. The `onFocus` is called every time the component gets focus (e.g. user clicks, tabs into it...). The `focus` method is there in case you need to give focus to the component programmatically.
Now, your `focus` method is more of an `onFocus` handler. Moreover, it's not meant for public use, I assume - you'd do that by saving `ref` to this component in a different component and then calling methods on it. So you can mark it with the keyword [private](https://www.typescriptlang.org/docs/handbook/classes.html#understanding-private). Then there's the naming convention in Draft that private methods should be prefixed with an underscore. So your method would be `_onFocus`. I'm not saying you need to follow this convention in your project, but you should not be surprised by it when you get to working with the Draft codebase ;)

src/actions/IAction.ts
```diff
@@ -0,0 +1,4 @@
+export interface IAction {
+  type: string;
```
vit-svoboda: Pretty much all of your interface properties can be marked [readonly](https://www.typescriptlang.org/docs/handbook/interfaces.html#readonly-properties). It doesn't make the objects immutable, as the types are deleted for runtime, but can sometimes point out errors in the code.

tsconfig.json
```diff
@@ -11,6 +11,7 @@
     "allowUnreachableCode": false,
     "allowUnusedLabels": false,
     "forceConsistentCasingInFileNames": true,
+    "allowSyntheticDefaultImports": true,
```
vit-svoboda: I thought you needed this when I looked at it the first time, but what is it actually good for? It fixes the `import React from 'react';` statements, but you use `import * as React from 'react';` everywhere anyway - and I'd stick with that way, since Draft imports React the same way.

src/components/todo-list/Input.tsx
```diff
@@ -44,11 +44,23 @@ export class Input extends PureComponent {
...
+    this.setState({
+      isFocused: true,
+    });
+  };
+
+  private _onBlur = (): void => {
+    this.setState({
+      isFocused: false,
+    });
+  };
```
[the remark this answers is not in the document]
matus12: tslint

I thought so, but when I put it above render, tslint is not happy about that. I guess render should not be private.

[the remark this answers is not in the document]
vit-svoboda: You're absolutely right, render has to be public and our [conventions](https://kentico.atlassian.net/wiki/spaces/KCD/pages/211791825/Complex+all-in-one+example) rely on the private methods not being marked `private`. So in your case, keep them private ('coz they are) and below `render` and be prepared for the Draft way, that's different a bit. I'm sorry about the confusion.

src/components/todo-list/AddedItem.tsx
```diff
@@ -26,10 +26,10 @@ export class AddedItem extends React.PureComponent<IAddedItemCallbackProps, ISta
     });
   }
 
-  changeItemText = (event: React.FormEvent<HTMLInputElement>): void => {
+  changeItemText = ({ currentTarget: { value } }: React.FormEvent<HTMLInputElement>): void => {
```
vit-svoboda: This and the `addItem` also belong to that underscore-prefixed private method category. Now are your components inconsistent.

src/components/todo-list/EditedItem.tsx
```diff
@@ -39,10 +39,10 @@ export class EditedItem extends React.PureComponent<IEditedItemCallbackProps & I
     };
   }
 
-  changeItemText = (event: React.FormEvent<HTMLInputElement>): void => {
+  changeItemText = ({ currentTarget: { value } }: React.FormEvent<HTMLInputElement>): void => {
```
vit-svoboda: Same goes for this and the `saveItem`.

PR #98, opened 2017-12-07
Feature/task 3

RadoslavK:


src/containers/ListContainer.jsx
```diff
@@ -0,0 +1,44 @@
...
+const mapStateToProps = state => {
+  const items = [];
+  state.list.items.forEach(item => items.push(item));
+
+  return {
+    items,
+  };
+};
+
+export const ListContainer = connect(
```
Neytus: It's not necessary to name this container with the `Container` suffix.

src/actions/actionCreatorsWithDependency.js
```diff
@@ -0,0 +1,13 @@
+import { ITEM_CREATED } from '../actionTypes';
+
+export const addNewItem = (createNewId, text) => {
```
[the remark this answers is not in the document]
PetrSvirak: @Neytus You really need to state why is that so, here. It might not be necessary for every review you will do, however, you basically have multiple ways to write a code that works, since you vote for a very specific one, reasons are definitely required unless you are absolutely sure everyone knows them ([but then, why would they choose another way to write the code?](https://memegenerator.net/img/images/600x600/16888855/clever-nigga.jpg)).

test/App.test.js
```diff
@@ -1,8 +1,21 @@
 import React from 'react';
```
Neytus: This test passes, however, there is a warning thrown. You can solve this by importing the `raf/polyfill` package. Make sure to import it above the import of `React`.

src/reducers/list/items/items.js
```diff
@@ -0,0 +1,75 @@
...
+  const newItem = new ListItem({
+    id,
+    text,
+  });
+
+  return state.set(id, newItem);
+};
+
+const deleteItem = (state, action) => {
+  const { itemId } = action.payload;
```
Neytus: You don't need to change anything, just think about cases where destructuring could lead to a more verbose code - this place is a nice example: the method can be written as a one-line lambda function (inlining).

src/containers/Items.jsx
```diff
@@ -0,0 +1,17 @@
...
+
+const mapStateToProps = state => {
+  const items = state.list.items.valueSeq().toArray();
+
+  return {
+    items,
+  };
+};
+
+const mapStateToPropsMoized = moize(mapStateToProps);
```
Neytus: You miss unit tests for memoization - it's better to be sure that it works as intended. In order to test it correctly, you need to move memoization away from this file.
As you could see in the implementation of `shouldDispatch` in `PureComponent`, it checks the first level of properties (in your case `items`) on whether they changed. So there's no need to memoize the whole props object. Just make sure you memoize the values of the properties if you don't just pass existing parts of the state to them.

src/containers/Items.jsx
```diff
@@ -0,0 +1,17 @@
+import { connect } from 'react-redux';
+import moize from 'moize';
+import { Items as ItemsComponent } from '../components/Items';
+
+const mapStateToProps = state => {
+  const items = state.list.items.valueSeq().toArray();
```
Neytus: You could achieve better performance by passing only the item IDs to the `List`. Then the `ListItem` component would be wrapped in a container that retrieves the item by given ID from the store.

PetrSvirak: Be more specific about the reasons why there could be better performance (in general - since I believe there will be no measurable effect in this small scale)

Neytus: Your memoization is now on the level of the state of your app - every time the state changes, you render the whole list, even though the change from the previous state could be just a little one. That's probably not how you want your app to work. 
Instead, if you would memoize on the level of a single item, then the only re-rendered part of your app would be the one corresponding to that specific item. Even though in this case it will not make a big difference, you could just pass the id of a specific ListItem from your `Items` component and then memoize either on the level of obtaining that single node or on creating a view model for that node. If you would have any further question then just ask.

src/actions/actionCreatorsWithDependency.js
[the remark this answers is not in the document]
Neytus: Just to show you some context as to why should you utilize [factory function](https://www.youtube.com/watch?v=ImwrezYhw4w): As Petr said, you have multiple ways how to write the code that works. However, in this case, we know for sure that we need dependency injection regarding  `createNewId` - without it, our function would not be testable at all. 

Concerning your previous `binding` approach, I have found out that it is very easy to break by mistakenly adding another attribute like that: 
```
const addNewItemFactory = () => _addNewItem.bind(null, createNewId, '12345');
```
Then the text attribute would be hardcoded as well. All in all, by adding more attributes with this approach, your code would become less readable and more prone to silly mistakes like mine above.

On the other hand, factory function has its dependencies clearly stated, is easy to test by just mocking the dependencies, and later on (Typescript), even harder to break as well. :-)

PR #99, opened 2017-12-17
Feature/task 4

LindaLangerova:


src/components/List.tsx
```diff
@@ -0,0 +1,36 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import { Seq } from 'immutable';
+import { ListItem } from '../containers/ListItem';
+import { ItemId } from '../models/ItemId';
+
+interface  IListDataProps {
+  ids: Seq.Indexed<string>;
```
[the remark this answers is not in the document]
LindaLangerova: Thanks! I totally overlooked this, it's ItemIds :)

src/components/List.tsx
```diff
@@ -0,0 +1,36 @@
...
+        </div>
+      )}
+    </div>
+  );
+};
+
+List.displayName = 'List';
+
+List.propTypes = {
+  ids: PropTypes.object.isRequired,
```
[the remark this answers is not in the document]
LindaLangerova: Nice, I will use that ^_^

src/components/AddNewItem.tsx
```diff
@@ -26,23 +37,24 @@ export class AddNewItem extends PureComponent {
...
             type="text"
-            onChange={this.onChange}
+            onChange={this._onChange}
             value={input}
             placeholder="Type new item name..."
           />
           <div className="input-group-btn">
             <button
-              data-balloon={isEmpty ? "Item name mustn't be empty" : null}
+              data-balloon={isEmpty ? 'Item name mustn\'t be empty' : null}
```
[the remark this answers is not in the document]
PetrSvirak: Actually, since `data-bolloon` reacts similarly to any falsy value, I would like to see `&&` operator here 😉 But otherwise, I like it as well 😃 👍

src/components/List.tsx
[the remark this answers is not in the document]
PetrSvirak: I definitely agree with replacing `PropTypes.object`, however, immutable prop-types AFAIK do not have type definitions, so TypeScript cannot validate if my (immutable) prop-types are `undefined`…

test/item.test.js
```diff
@@ -0,0 +1,10 @@
+import Immutable from 'immutable';
+import { Item, defaultItem } from '../src/models/Item.ts';
+
+describe('Item', () => {
+  it('creates immutable object', () => {
+    const item = new Item(defaultItem);
+
+    expect(item).toBeInstanceOf(Immutable.Record);
```
PetrSvirak: That is exactly what I meant regards @Neytus's comment (above) - I see no purpose in this test. One might test this object really is immutable, however, the whole point of the immutable library is that you don't have to. Also, only the reducers should ever create a new instance of the record, all other parts of the application should always work with read-only interface `IItem`. I'm open to any value test like this might bring, however, I presently see none.

test/newItemText.test.js
```diff
@@ -1,13 +1,16 @@
...
-  addItem,
+  addItemFactory,
   deleteItem,
   updateNewItemText,
-} from '../../actions/actionCreators';
-import { generateId } from '../../utils/generateId';
+} from '../src/actions/actionCreators.ts';
+
+function mockId() {
+  return '2';
```
PetrSvirak: This very function almost screams "lambda" 😆

src/actions/actionCreators.ts
```diff
@@ -0,0 +1,55 @@
...
+  return {
+    type: actionTypes.ADD_ITEM,
+    payload: {
+      text,
+      id,
+    },
+  };
+};
+
+export function addItemFactory (idFunc: () => string, text: string) {
```
PetrSvirak: First of all, I don't like factory being placed in this very file. Move it to a stand-alone file, please, so it does not mix up with simple actions it is surrounded with. Feel free to take the `addItem` with it and even inline the simpler action creator into the factory. [TBC](https://github.com/KenticoAcademy/kentico-onboarding-js/pull/99/files#diff-590cfc6cb81be063d07de0cf323caba9R13)

src/containers/AddNewItem.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { IAppState } from '../stores/IAppState';
+import { generateId } from '../utils/generateId';
+import { addItemFactory, updateNewItemText } from '../actions/actionCreators';
+
+const mapStateToProps = ({ items: { newItemText} }: IAppState): IAddNewItemDataProps => ({
+  newItemText,
+});
+
+const mapDispatchToProps = (dispatch: Dispatch<IAppState>): IAddNewItemCallbackProps => ({
+  onAdd: (value: string) => dispatch(addItemFactory(generateId, value)),
```
PetrSvirak: Second ([CF](https://github.com/KenticoAcademy/kentico-onboarding-js/pull/99/files#diff-168539a9a821bc304fb41999c583f3adR15)), you inject your dependency along the parameters your function actually accepts. JavaScript perfectly allows you to merge factory and its result, however, it makes code less readable, requires you to write arguments in a specific order is not done here. 

Split the function in two - first function fill only accept dependencies and return the second, that will only accept arguments (the second function will be the result of the first one, thus it will have to all dependencies through closure).

src/actions/actionCreators.ts
PetrSvirak: Third ([CF](https://github.com/KenticoAcademy/kentico-onboarding-js/pull/99/files#diff-590cfc6cb81be063d07de0cf323caba9R13)), you should have your injection done on a single place for each group of dependent function (you will leverage this in Task 5), so I suggest creating "a public interface" for your actions within the application - create an `index.ts` in this folder and only import that wherever you need an action across the application (tests are the only exception here) - you can do your dependency injection there. 
Note: use `export from` to re-export basic action.

src/components/List.tsx
```diff
@@ -1,15 +1,14 @@
 import * as React from 'react';
-import { Seq } from 'immutable';
+import * as Immutable from 'immutable';
 import { ListItem } from '../containers/ListItem';
 import { ItemId } from '../models/ItemId';
-const ImmutablePropTypes = require('react-immutable-proptypes');
+import ImmutablePropTypes = require('react-immutable-proptypes');
```
PetrSvirak: You can use `import * as ImmutablePropTypes from 'react-immutable-proptypes';` as you would before TypeScript 😉

src/models/BaseRecord.ts
```diff
@@ -1,7 +1,7 @@
 import { Record } from 'immutable';
 
-export const BaseRecord = (defaultValues: {}) => class<T> extends Record(defaultValues){
-  with(update: Partial<T>) {
+export const BaseRecord = (defaultValues: {}) => class extends Record(defaultValues){
+  with(update: Partial<this>) {
     return this.merge(update) as this;
   }
```
milan009: Oh and by the way - Immutable Record supports one additional parameter - a string specifying the name of that Record. This name is used when converting the Record to text (like when passing it to an Error - [tiny bit of reading about it here](https://medium.com/@ahmadbamieh/immutablejs-records-bd369137da06)). Your `BaseRecord` should utilize this feature as well.

src/models/Item.ts
```diff
@@ -17,7 +17,7 @@ export const defaultItem: IItem = {
   textUpdate: '',
 };
 
-export class Item extends BaseRecord(defaultItem)<Item> implements IItem {
+export class Item extends BaseRecord(defaultItem) implements IItem {
```
milan009: If you make your BaseRecord generic again, you will have to revert this change 😉

milan009: Should you make the Record generic, try moving your constructor from `Item` model into `BaseRecord` - this way you can reuse that piece of code whenever you create additional derived record type.

PR #104, opened 2018-02-23
Feature/task 2

vaclavm:


src/models/toDoItem.js
```diff
@@ -0,0 +1,7 @@
+import { Record } from 'immutable';
+
+export const ToDoItem = Record({
+  key: undefined,
+  value: '',
+  bullet: undefined,
+});
```
milan009: When creating a custom Record, you can pass a second parameter that specifies the name of the Record - this name is used while converting the Record to a string - either in logging or when passed as a parameter to Error [[link]](https://medium.com/@ahmadbamieh/immutablejs-records-bd369137da06#1d44). You should always name your Records for the sake of easier debugging and logging :wink:

Additionally, you could extract the object literal that specifies the default values for the record into a separate const and than pass that one to the constructor. It'd make the code a little more readable and easier to modify/extend.

Finally, I don't really like that your Record includes the `bullet` property - the main reason you are introducing Immutable Records is to enforce the immutability of your React state, your Records are therefore used mainly in the state. Yet, you never ever set nor read the bullet value in the state - every single item within the store always contains the default `undefined` value for that property. This just doesn't feel right.

This related to another thing - I see that you've started using the Record in PropTypes of corresponding components - I guess the goal was to have a single representation of Item to store and pass around, yet that is not necessary, sometimes not even desirable. For example, now your `ListItem` component requires its `item` prop to have a `value` property (since it has to be your Record) while never utilizing it in any way - this is the exact way we've been talking about and that you've convinced me would impose extra responsibility for that component 😉

PetrSvirak: /> Finally, I don't really like that your Record includes the bullet property - the main reason you are introducing Immutable Records is to enforce the immutability of your React state, your Records are therefore used mainly in the state. Yet, you never ever set nor read the bullet value in the state - every single item within the store always contains the default undefined value for that property. This just doesn't feel right.

That is where you are wrong, @milan009.  Immutability is essential for both state and props since pure components rely on their immutability. It is actually more important for props since they are shared across multiple components while the state is more "local". Now, passing the bullet down individually is an option. However, I would consider having two Records (models) where one is created (and properly memoized) based on another - this is where we will be heading to in Task 3 anyway.

Also, @milan009: When you happen to describe multiple tasks that need to be performed in order to fix something, leave each in a standalone comment, so one can easily mark individual tasks as done...

What @milan009 overlooked or you did not fix is that:

PetrSvirak: The default data of the record should be stored in a standalone variable.

PetrSvirak: You mustn't use `undefined` in default data - these values are supposed to possess default but valid value - i.e. empty string or zero, ... - never a `null` or `undefined` which represent missing value in one way or another.

PetrSvirak: The `name` in the Record class factory is supposed to describe the type rather than individual instances (and you use it this way anyway 😉), please, move the name to the record factory call.

PetrSvirak:
After some discussion with @vaclavm, we decided to go with a thorough review in the following task since many parts of troubling code here have been already changed in the task. Closing & merging now…

PR #105, opened 2018-02-23
Feature/task 5

kubmir:


src/components/inactiveNote/InactiveNote.tsx
```diff
@@ -0,0 +1,39 @@
...
+    readonly communicationError: string;
+    readonly failedAction: FailedAction;
+  };
+  readonly number: number;
+}
+
+type InactiveNoteProps = IInactiveNoteDataProps;
+
+const InactiveNote: React.StatelessComponent<InactiveNoteProps> = (props: InactiveNoteProps): JSX.Element => {
+  const InactiveNoteComponent = props.note.communicationError === ''
```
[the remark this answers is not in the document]
PetrSvirak: You check whether `props.note.communicationError === ''` – it is more usual to check against a `falsy` value (`props.note.communicationError` being `null` or `undefined` (mind you: reducers technically all only `null`)) - are you sure `''` will always mean "no error" – what if server returns an empty error message?

src/components/addNote/AddNote.tsx
```diff
@@ -1,16 +1,17 @@
 import * as React from 'react';
 import * as PropTypes from 'prop-types';
-import { NoteErrorMessage } from './NoteErrorMessage';
-import { AddNoteInput } from '../containers-redux/AddNoteInput';
-import { isNoteValid } from '../utils/isNoteValid';
+import { AddNoteInput } from '../../containers-redux/addNote/AddNoteInput';
```
PetrSvirak: It is a bit weird that you use a different approach to the`NonEmptyInput` in `AddNote` and `NoteEditor`, please unify them.

src/models/IServerNote.ts
```diff
@@ -0,0 +1,6 @@
+export interface IServerNote {
```
PetrSvirak: I'm really considering whether not merge this file into the `Note.ts` since it is likely you will use it with `INote` only…

src/models/Note.ts
```diff
@@ -1,21 +1,31 @@
 import { Record } from 'immutable';
+import { FailedAction } from '../enums/failedAction';
```
PetrSvirak: The filename does not match the main export name.

test/testUtils/mocks.ts
```diff
@@ -0,0 +1,48 @@
...
+export const mockResolvedRequest = (responseBody: string): Promise<IMockedResponse> =>
+  Promise.resolve(mockResponse(200, true, responseBody));
+
+export const mockRejectedRequest = (): Promise<IMockedResponse> =>
+  Promise.reject(mockResponse(200, true));
+
+export const FALSE_INITIAL_STATE = false;
+export const TRUE_INITIAL_STATE = true;
+
+export const mockStoreState = (): IStoreState =>
```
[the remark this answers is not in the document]
kubmir: I use it in actions -> thunkFactories tests. 

```javascript
return deleteNoteFactory(deleteNoteDependencies)(NOTE_TO_DELETE_ID)(dispatch, () => mockStoreState(), null)
```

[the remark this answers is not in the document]
PetrSvirak: I know that is why I think it can be replaced with the root reducer call. Otherwise, any change to your redux store means a potential breakdown of these tests.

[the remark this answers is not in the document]
kubmir: Thanks, you are right. I will change it.

src/reducers/errors/errors.ts
```diff
@@ -0,0 +1,54 @@
...
+    id: errorId,
+    errorDescription,
+    failedAction,
+  });
+
+  return state
+    .set(errorId, errorToAdd);
+};
+
+const removeError = (state: OrderedMap<Guid, ApplicationError>, payload: { errorId: Guid }): OrderedMap<Guid, ApplicationError> =>
```
PetrSvirak: This method is quite useless 😛

src/reducers/notes/list/note/note.ts
```diff
@@ -0,0 +1,110 @@
+import { IAction } from '../../../../actions/IAction';
```
PetrSvirak: Placing the `note` reducer into a folder of the same name makes it inconsistent with other reducers – maybe having a folder `notes` with `notes.ts` and `note.ts` in it would be better.

PR #107, opened 2018-03-01
on-boarding improvement task-0 CI

kubmir:
Hello Peťo,

I did some changes in the description of task-0 to extend it for continuous integration setup. If you think the description should be more accurate I will re-work it next week.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,16 @@ ESlint and TSlint are already set-up for you, so you will see all the errors and
 
 ## Task 0
 Update your upstream solutions branch with the changes that might have happened on upstream master-v2: After cloning your repository, create `develop` and then `feature/task-0` branch (as described above).
-Merge your `master-v2` branch - that already includes all changes from `upstream/master-v2` - into task-0 branch and submit new Pull request to your solutions branch from task-0 branch.
+Merge your `master-v2` branch - that already includes all changes from `upstream/master-v2` - into task-0 branch. 
+
+Your next step will be to set up a [continuous integration](https://www.atlassian.com/continuous-delivery/continuous-integration-intro) in your repository. 
+- each push of commits should trigger tests
+- each push to the _master_ branch should trigger release
+- status of a build should be visible as a GitHub badge in repository (search for "status badge _your CI name_")
+- you can choose any CI that is compatible with GitHub and has free license/plan to be used by company's employee - for example [Travis CI](https://travis-ci.org/), [AppVeyor](https://www.appveyor.com/) or [Circle CI](https://circleci.com/)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -64,7 +64,16 @@ ESlint and TSlint are already set-up for you, so you will see all the errors and
 
 ## Task 0
 Update your upstream solutions branch with the changes that might have happened on upstream master-v2: After cloning your repository, create `develop` and then `feature/task-0` branch (as described above).
-Merge your `master-v2` branch - that already includes all changes from `upstream/master-v2` - into task-0 branch and submit new Pull request to your solutions branch from task-0 branch.
+Merge your `master-v2` branch - that already includes all changes from `upstream/master-v2` - into task-0 branch. 
+
+Your next step will be to set up a [continuous integration](https://www.atlassian.com/continuous-delivery/continuous-integration-intro) in your repository. 
+- each push of commits should trigger tests
+- each push to the _master_ branch should trigger release
```
PetrSvirak: Do your repositories do this? I mean, running tests is one thing, but setting up release is quite another. There is also no link that would guide through this process. I'm not sure whether removing this bullet, especially since Azure takes some time to digest, or adding some links.

kubmir: I was not sure about the release when I was writing it.  However, I agree Azure takes some time to digest so I will remove this bullet from task-0 and release will remain a part of publish task - same as now. 

I also noticed a problem with [exceeded storage](https://help.appveyor.com/discussions/problems/10253-maximum-allowed-artifact-storage-size-of-1000-mb-will-be-exceeded) limit on Appveyor CI. If CI will be a part of tasks 0 more people can have same issue - I suggest Travis be marked as a preferred provider. What do you think about it?

Should I add link also for setting up CI on Travis as an example or you meant it just for release?

README.md
```diff
@@ -68,7 +68,6 @@ Merge your `master-v2` branch - that already includes all changes from `upstream
 
 Your next step will be to set up a [continuous integration](https://www.atlassian.com/continuous-delivery/continuous-integration-intro) in your repository. 
 - each push of commits should trigger tests
-- each push to the _master_ branch should trigger release
 - status of a build should be visible as a GitHub badge in repository (search for "status badge _your CI name_")
 - you can choose any CI that is compatible with GitHub and has free license/plan to be used by company's employee - for example [Travis CI](https://travis-ci.org/), [AppVeyor](https://www.appveyor.com/) or [Circle CI](https://circleci.com/)
```
PetrSvirak: The term "company's employees" seems to be a bit misleading legally (seems almost like there was a list (our) company approved CIs). Try "… and has a free license/plan for commercial use…"

PR #108, opened 2018-03-12
Feature/task 5

RadoslavK:


test/actions/thunk/fetchItems.test.ts
```diff
@@ -0,0 +1,69 @@
...
+      .catch(fakeFunction);
+  });
+
+  it('requestItems and fetchFailed actions', () => {
+    const expectedActions = [
+      'requestItems',
+      'fetchFailed',
+    ];
+    const httpClient = httpClientFailure;
+    const requestItems = jest.fn(() => expectedActions[0]);
```
Suzii: It is completely useless to have this function as mock. You never use it as a mock. The only purpose it serves here is that you are 'losing' type checking. However, once you stop injecting the base actions to thunk action creators, you will have to come up with another way for checking that right action has been dispatched. I suggest changing `getFirstArgumentOfCalls` helper to `getDispatchedActionType` and return string constant and check those against constants in `actionTypes.ts` 😉

test/actions/thunk/thunkTestsUtils.ts
```diff
@@ -0,0 +1,37 @@
...
+export const httpClientFailure: IHttpClient = ({
+  get: rejector,
+  delete: rejector,
+  post: rejector,
+  patch: rejector,
+  put: rejector,
+});
+
+export const fakeFunction = jest.fn();
+export const dispatch = jest.fn();
```
Suzii: It is not a good idea to have a global disptach mock for asynchronous tests, as they may influence one another in some cases. Please create a dedicated mock per test case (you can just change this line to a `getDispatchMock()` factory).

test/actions/thunk/fetchItems.test.ts
Suzii: Applies to all tests of thunks.

src/models/classes/AxiosHttpClient.ts
```diff
@@ -0,0 +1,34 @@
+import axios, { AxiosResponse } from 'axios';
+import { IHttpClient } from '../interfaces/IHttpClient';
+
+const getData = <T>(res: AxiosResponse<T>) =>
+  res.data;
+
+class AxiosHttpClient implements IHttpClient {
```
Suzii: It does not need to be a class at all. Just declare a variable of type `IHttpClient` and create a simple object literal conforming to this interface.

src/components/App.tsx
```diff
@@ -1,40 +1,36 @@
 import '../styles/sticky-footer.css';
 import * as React from 'react';
-import { PureComponent } from 'react';
 import { List } from '../containers/List';
 
-export class App extends PureComponent {
-  static displayName = 'App';
+const App: React.SFC = () =>
```
PetrSvirak: Don't use the [`SFC`](https://github.com/Microsoft/TypeScript/blob/24e35cd940be25853c16334c6a2143ee33a4b580/tests/lib/react.d.ts#L199) shortcut even if it's [SFC](https://www.urbandictionary.com/define.php?term=SFC#5539742), please – it might easily be interpreted as [SFC](https://www.urbandictionary.com/define.php?term=SFC#9781080)

src/reducers/list/itemsSyncInfo/index.ts
```diff
@@ -23,82 +23,64 @@ import { SyncOperation } from '../../../models/enums/SyncOperation';
...
+
+      return itemSyncInfo
+        ? state.update(
+          itemSyncInfo.id,
+          new ItemSyncInfo({
+            id: itemSyncInfo.id,
+          }),
+          originalSyncInfo => syncInfo(originalSyncInfo, action)
+        )
+        : state;
```
PetrSvirak: This is weird. If there is one of the actions that does not possess `itemSyncInfo`, it should not be in the case. If there are some actions that don't have `itemSyncInfo` based on some circumstances, these actions should be split in two and the ones without `itemSyncInfo` shouldn't further participate in this reducer. It seems that you dispatch a more-or-less omni-action or two 😉

PR #111, opened 2018-03-21
Feature/task 1

butcher-07:
Review improvements added, I hope this time it will be much better and thanks for very deep reviews guys.

### DIFF
--- a/src/components/CreateListItem.jsx
+++ b/src/components/CreateListItem.jsx
@@ -1,5 +1,6 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
+import { enableUniqueIds } from 'react-html-id';
 
 export class CreateListItem extends PureComponent {
   static displayName = 'CreateListItem';
@@ -8,17 +9,22 @@ export class CreateListItem extends PureComponent {
     onSubmit: PropTypes.func.isRequired,
   };
 
-  state = {
-    newItemText: '',
-  };
+  constructor(props) {
+    super(props);
+    this.state = {
+      newItemText: '',
+    };
 
-  _inputChange = (e) => {
+    enableUniqueIds(this);
+  }
+
+  _updateItemText = (e) => {
     this.setState({
       newItemText: e.target.value,
     });
   };
 
-  _inputSubmit = (e) => {
+  _submitItemText = (e) => {
     e.preventDefault();
     this.props.onSubmit(this.state.newItemText);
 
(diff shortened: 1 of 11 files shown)


src/components/CreateListItem.jsx
```diff
@@ -0,0 +1,47 @@
...
+      newItemText: '',
+    });
+  };
+
+  render() {
+    return (
+      <li className="list-group-item">
+        <span className="form-group">
+          <label htmlFor={this.nextUniqueId()} />
+          <input type="text" className="form-control" id={this.lastUniqueId()} value={this.state.newItemText} onChange={this._updateItemText} />
```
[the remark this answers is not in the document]
PetrSvirak: \[OPTIONAL\]: You might eventually want to extract your first reusable/UI/generalized component here – whether you use `label` or `aria-`, you will always require same standards from the input in the `CreateListItem` and `ListItemEditor`. It is also a good way to ensure that the `label`'s text is not forgotten or the `react-html-id` is not scattered all over the place.

src/components/CreateListItem.jsx
```diff
@@ -0,0 +1,47 @@
...
+    this.setState({
+      newItemText: '',
+    });
+  };
+
+  render() {
+    return (
+      <li className="list-group-item">
+        <span className="form-group">
+          <label htmlFor={this.nextUniqueId()} />
```
PetrSvirak: I still believe the `label` itself should [not be empty](https://achecker.ca/checker/suggestion.php?id=188). It might need to [be hidden](https://www.w3.org/WAI/tutorials/forms/labels/#hiding-the-label-element), of course.

butcher-07:
So last changes added, will see if it's okay now...

src/components/ListItemDisplay.jsx
```diff
@@ -5,11 +5,10 @@ export class ListItemDisplay extends PureComponent {
   static displayName = 'ListItemDisplay';
 
   static propTypes = {
-    onClick: PropTypes.func.isRequired,
     text: PropTypes.string.isRequired,
   };
 
   render() {
-    return <span onClick={this.props.onClick}>{this.props.text}</span>;
+    return <span>{this.props.text}</span>;
```
PetrSvirak: Nothing wrong with this `span`, however, you might love to use [fragments](https://reactjs.org/blog/2017/11/28/react-v16.2.0-fragment-support.html) from React 16 in similar situations 😃

PR #114, opened 2018-05-28
Feature/task 2

butcher-07:
Task 2 introducing Immutability

### DIFF
--- a/src/components/List.jsx
+++ b/src/components/List.jsx
@@ -1,56 +1,64 @@
 import React, { PureComponent } from 'react';
+import { OrderedMap } from 'immutable';
 import assignment from './../../assignment.gif';
 
-import { UniqueIdentifier } from '../utils/UniqueIdentifier.jsx';
 import { ListItem } from './ListItem.jsx';
 import { CreateListItem } from './CreateListItem';
+import { ItemData } from '../models/ListItemData.jsx';
+import { UniqueIdentifier } from '../utils/UniqueIdentifier';
 
 export class List extends PureComponent {
   static displayName = 'List';
 
   state = {
-    items: [],
+    items: OrderedMap(),
   };
 
   _addItem = itemText => {
-    const newItem = {
-      text: itemText,
+    const item = new ItemData({
       id: UniqueIdentifier.generateUniqueId(),
-    };
-
+      text: itemText,
+    });
     this.setState(prevState => ({
-      items: [
-        ...prevState.items,
-        newItem,
-      ],
+      items: prevState.items.set(item.id, item),
     }));
   };
 
-  _deleteItem = (itemId) =>
-    this.setState(prevState => ({
-      items: prevState.items.filter(item => item.id !== itemId),
+  _deleteItem = (itemId) => {
+    return this.setState(prevState => ({
+      items: prevState.items.delete(itemId),
     }));
+  };
 
-  _updateItem = (itemId, newText) =>
+  _updateItemText = (itemId, newText) => {
+    const item = new ItemData({
+      id: itemId,
+      text: newText,
+    });
     this.setState(prevState => ({
-      items: prevState.items.map(item => (
-        item.id !== itemId
-          ? item
-          : {
-            ...item,
-            text: newText,
-          })),
+      items: prevState.items.set(itemId, item),
     }));
+  };
+
+  _toggleItemEdit = (itemId, onlyTurnOn = false) => {
+    if (!onlyTurnOn || !this.state.items.get(itemId).isEdited) {
+      this.setState(prevState => ({
+        items: prevState.items.update(itemId, item => item.update('isEdited', value => !value)),
+      }));
+    }
+  };
 
   render() {
-    const listItems = this.state.items.map((item, index) =>
+    const listItems = this.state.items.valueSeq().map((item, index) =>
       <ListItem
         id={item.id}
         key={item.id}
         number={index + 1}
         text={item.text}
-        onChange={this._updateItem}
+        onChange={this._updateItemText}
         onDelete={this._deleteItem}
+        onToggle={this._toggleItemEdit}
+        inEditMode={item.isEdited}
       />);
 
     return (
--- a/src/components/ListItem.jsx
+++ b/src/components/ListItem.jsx
@@ -10,33 +10,20 @@ export class ListItem extends PureComponent {
   static propTypes = {
     text: PropTypes.string,
     id: PropTypes.string,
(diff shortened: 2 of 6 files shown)


src/components/List.jsx
```diff
@@ -1,56 +1,64 @@
...
-            text: newText,
-          })),
+      items: prevState.items.set(itemId, item),
     }));
+  };
+
+  _toggleItemEdit = (itemId, onlyTurnOn = false) => {
+    if (!onlyTurnOn || !this.state.items.get(itemId).isEdited) {
+      this.setState(prevState => ({
+        items: prevState.items.update(itemId, item => item.update('isEdited', value => !value)),
```
PetrSvirak: While using `update('isEdited', …)` is quite JavaScriptish, I believe there is a bit cleaner way. You can leave it as it is until TypeScript forces you to [change this line](https://imgflip.com/i/2bfc1a) or try to change it right away.

<details>
  <summary>Spoiler</summary>

   ````
items: prevState.items.update(itemId, item => item.merge({ isEdited: !item.isEdited }))
   ````

It is equaly extensible to you current code in terms of properties of the `ItemData`, however, it enables type-check in the future and extension of the number of amended properties "right now" (I know, I know, it is very unlikely for this method to change how many properties of `ItemData` it changes)
</details>

src/components/List.jsx
```diff
@@ -1,56 +1,64 @@
...
+      items: prevState.items.delete(itemId),
     }));
+  };
 
-  _updateItem = (itemId, newText) =>
+  _updateItemText = (itemId, newText) => {
+    const item = new ItemData({
+      id: itemId,
+      text: newText,
+    });
```
PetrSvirak: \[Note]: Please attend after solving the comment at [line 46](https://github.com/KenticoAcademy/kentico-onboarding-js/pull/114/commits/02ebcbd86f41c515dbed6504b89133075d0d5d3a#diff-5ddcbc3416ad10b53b15941fff31e4e0R46).

Could you rewrite this method without explicitly creating a new object? It is an update, after all…

<details>
  <summary>Spoiler</summary>

It is not only about me seeing the flow of life of the object. It effectively allows the `ItemData` to be freely extended while your implementation strips any other properties than `id` and `text`. Right now, it implicitly (rather than explicitly) resets the `isEdited` flag. It is nice of you not to misuse this "feature" of your code though 😃 

Moreover, `Immutable.js` provides [quite usual](https://imgflip.com/i/2bfahr) API to [manipulate both collections and records](https://imgflip.com/i/2bfb9k):
<details>
  <summary>Spoiler</summary>

   You can combine `update` that accepts a function with `merge` that provides more control over updated record (especially little later on, with TypeScript). There are also these `mergeIn` and `updateIn` methods that have some potential; on the other hand, it might affect clarity of your code if overused.

Something like following code will do the trick (while keeping the `update` method pretty fall-through):
````
_updateItemText = (itemId, newText) => this.setState(
  prevState => ({
    items: prevState
      .items
      .update(
        itemId,
        existingItem => existingItem.merge({ text: newText })),
  }));
````

</details>
</details>

File: src/components/List.jsx [hunk 3]
```diff
@@ -1,56 +1,64 @@
 import React, { PureComponent } from 'react';
+import { OrderedMap } from 'immutable';
 import assignment from './../../assignment.gif';
 
-import { UniqueIdentifier } from '../utils/UniqueIdentifier.jsx';
 import { ListItem } from './ListItem.jsx';
 import { CreateListItem } from './CreateListItem';
+import { ItemData } from '../models/ListItemData.jsx';
```
PetrSvirak: There is not JSX code in the file, so `.js` the extension is preferable…

PetrSvirak: Your files should be named after most significant (if not only) "object" you export from them - `ItemData` is what I believe to be the most significant object (function) in this case.

src/components/List.jsx
```diff
@@ -1,56 +1,64 @@
...
+import { UniqueIdentifier } from '../utils/UniqueIdentifier';
```
PetrSvirak: I must have missed the extension in this file's name last time 👼

src/components/List.jsx
```diff
@@ -1,56 +1,64 @@
...
 
 export class List extends PureComponent {
   static displayName = 'List';
 
   state = {
-    items: [],
+    items: OrderedMap(),
```
PetrSvirak: 👍

src/components/List.jsx
```diff
@@ -1,56 +1,64 @@
...
-      ],
+      items: prevState.items.set(item.id, item),
     }));
   };
 
-  _deleteItem = (itemId) =>
-    this.setState(prevState => ({
-      items: prevState.items.filter(item => item.id !== itemId),
+  _deleteItem = (itemId) => {
+    return this.setState(prevState => ({
```
PetrSvirak: Preferably just omit body when a function starts with a `return` statement.

src/components/ListItem.jsx
```diff
@@ -10,33 +10,20 @@ export class ListItem extends PureComponent {
   static propTypes = {
     text: PropTypes.string,
     id: PropTypes.string,
+    inEditMode: PropTypes.bool,
```
PetrSvirak: Wouldn't it be nice to pass whole `ItemData` down the `ListItem` so individual props don't mix up? 😇 There is even a nice `PropTypes`'s method that saves you edition of `propTypes` in the future…

src/components/ListItem.jsx
```diff
@@ -46,7 +33,7 @@ export class ListItem extends PureComponent {
         text={this.props.text}
         onDelete={this.props.onDelete}
         onChange={this._changeItem}
-        onCancel={this._toggleEditMode}
+        onCancel={this.props.onToggle}
```
PetrSvirak: Just a thing to know (I'm not saying it is the best thing to do, however:), it is possible to pass `props` from one component to another like this:
````
<ListItemEditor
  ...this.props
  onChange={this._changeItem}
/>);
````

or even
````
<ListItemEditor
  ...this.props.item
  ...this.props
  onChange={this._changeItem}
/>);
````
if You decided to follow my suggestion on [line 13](https://github.com/KenticoAcademy/kentico-onboarding-js/pull/114/commits/02ebcbd86f41c515dbed6504b89133075d0d5d3a#diff-484fc9de16fe7dbcecabdd06079150a7R13), but not change `ListItemEditor`.

It might save you same lines, however, it is always opened for discussion whether it improves readability in the particular case or not.

PR #124, opened 2018-07-13
Feature/task 1

HanaNavratilova:


src/components/ActiveItem.jsx
```diff
@@ -0,0 +1,55 @@
...
+
+  saveInputValue = () => this.props.onSaveItem(this.state.text);
+
+  storeInputValue = e => this.setState({ text: e.target.value });
+
+  render() {
+    return (
+      <li className="list-group-item">
+        <div className="input-group col-md-8">
+          <span className="input-group-addon">
```
PetrSvirak: Not that it is nice - it is a kinda bootstrap hack actually - however, if you don't like the gray background of the item number, you can actually create a markup like this:
````
<span class="input-group-btn">
    <input value="2." type="button" class="btn btn-default" disabled="disabled">
</span>
````
It has its own disadventages though…

File: src/components/ActiveItem.jsx [hunk 2]
```diff
@@ -0,0 +1,55 @@
...
+            {
+              this.props.index + 1
+            }
+            .
+          </span>
+        <input className="form-control" type="text" value={this.state.text} onChange={this.storeInputValue} />
+        <span className="input-group-btn">
+          <button className="btn btn-primary" type="submit" onClick={this.saveInputValue}>
```
PetrSvirak: You don't have the attributes wrapped, here. Simply reformat the file using `CTRL` + `ALT` + `L` or set WebStorm to reformat your file on save (you would have to use macros for that I'm afraid, but ask guys around, I believe Milan has this set up).

src/utils/guid.test.js
```diff
@@ -0,0 +1,21 @@
+import { guid } from './guid';
+
+const regex = RegExp(/^[a-fA-F0-9]{8}-([a-fA-F0-9]{4}-){3}[a-fA-F0-9]{12}$/);
```
PetrSvirak: That's much better :) I'm still not completely satisfied with your regex since uuid (or guid) is pretty [specific format and has a version encoded in it](https://tools.ietf.org/html/rfc4122#section-4.1.3). It is just a detail, so please, fix it in the following task.

src/utils/guid.test.js
```diff
@@ -0,0 +1,21 @@
...
+
+    expect(testedGuid)
+      .toMatch(regex);
+  });
+
+  it('is unique', () => {
+    const guidArray = Array(20).fill(undefined).map(() => guid());
+    const guidSet = new Set(guidArray);
+
+    expect(guidArray.length)
```
PetrSvirak: Preferably keep expects on a single line - it emulates natural sentense more closely this way. But that is again a tiny detail - it would be great to change it in the next task.

PR #126, opened 2018-07-19
Feature/task 2

HanaNavratilova:


src/components/List.jsx
```diff
@@ -16,24 +16,26 @@ export class List extends PureComponent {
...
-    };
-  });
+  _deleteItem = id => this.setState(prevState => ({
+    list: prevState.list.delete(id)
+  }));
 
 
   _addItem = text => this.setState(prevState => {
     const id = guid();
-    const newItem = new ImmutableItem({ id, text });
```
PetrSvirak: This did not help too much in this case. Either preserve the original or be consistent with the `_editItem` (and place `guid()` into the object you would return directly).

src/components/List.jsx
```diff
@@ -1,60 +1,43 @@
...
     return {
-      list: [
-        ...prevState.list,
-        newItem
-      ]
+      list: prevState
+        .list
+        .set(
+          id,
+          new ListItem({
```
PetrSvirak: Since you need to have a method body anyway, extract the constructor call to a stand-alone variable - `CTRL` + `ALT` + `V` (and then `ALT` + `C`) should help you with that, provided you select everything from the `new` to the `})`. I believe it is appropriate not to request changes on this single one and see this approach in the next task.

PR #127, opened 2018-07-23
Feature/task 1

petrareb:


src/components/ItemToAdd.jsx
```diff
@@ -0,0 +1,68 @@
...
+    };
+  }
+
+  _changedTextInput = (event) => {
+    this.setState({
+      text: event.target.value
+    });
+  };
+
+  _onAddClick = () => {
```
vaclavm: Naming :)

src/components/TodoList.jsx
```diff
@@ -76,29 +68,27 @@ export class TodoList extends React.PureComponent {
...
+        return (
           <TodoListItem
-            itemProps={
-              <ItemProperties
-                id={item.id}
-                text={item.text}
-                index={i + 1}
-              />
-            }
-            key={item.id}
```
vaclavm: By removing the key from the component, you have created error in console:
image
If you iterate through array and create html components, every one should have a key attribute.

src/components/AddItem.jsx
```diff
@@ -4,26 +4,20 @@ import { isValidText } from '../utils/validateText';
 import classNames from 'classnames';
 
 
-export class ItemToAdd extends React.PureComponent {
+export class AddItem extends React.PureComponent {
```
vaclavm: AddItem is more like a functionality than a component name. Better would be NewItem or ItemToAdd (your previous version).

src/components/AddItem.jsx
```diff
@@ -4,26 +4,20 @@ import { isValidText } from '../utils/validateText';
...
-  _changedTextInput = (event) => {
+  _changedTextInput = (event) =>
     this.setState({
       text: event.target.value
     });
-  };
 
-  _onAddClick = () => {
+  _addNewItem = () => {
     if (this.state.text) {
```
vaclavm: I think you do not need this condition as you are not able to click od Add unless something is written in the field.

src/components/EditItem.jsx
```diff
@@ -69,7 +66,7 @@ export class EditItem extends React.PureComponent {
               type="button"
               name="itemToModifySaveButton"
               value="Save"
-              onClick={this._onSaveClick}
+              onClick={this._editItem}
               disabled={!isValidText(this.state.text)}
```
vaclavm: Instead of calling !isValidText(this.state.text) three times, create a const for it and pass the const ;-)

src/components/Page.jsx
```diff
@@ -3,7 +3,10 @@ import assignment from '../assignment.gif';
 
 import { TsComponent } from './TsComponent.tsx';
 
-export class List extends PureComponent {
+import { TodoList } from './TodoList.jsx';
+
+
+export class Page extends PureComponent {
```
vaclavm: Can be const and add displayName for it.

src/components/EditItem.jsx
```diff
@@ -19,29 +19,26 @@ export class EditItem extends React.PureComponent {
...
 
-  _onDeleteClick = () => {
-    this.props.onDeleteClick(this.props.itemProps.props.id);
-  };
+  _deleteItem = () =>
+    this.props.onDeleteClick(this.props.item.id);
 
-  _onSaveClick = () => {
+  _editItem = () => {
     if (this.state.text) {
```
vaclavm: I think this condition is not needed as you don't allow to click on Save unless some text is in the input.

PR #129, opened 2018-07-26
Feature/task 2

petrareb:


src/utils/generateId.js
```diff
@@ -1,3 +1,3 @@
-import * as generateUuidV1 from 'uuid/v1';
+import v4 from 'uuid/v4';
```
vaclavm: You should use proper naming even when importing functions from 3rd party libraries. It seems useless in this case, but imagine you have a file with 100+ lines of code with line `const v4 = v4();`. Would you understand what it is?
Maybe it would be better to see this line: `cons uuidV4 = generateUuidV4();`. Soooooo use naming import ;-)

src/utils/validateText.js
```diff
@@ -1 +1,2 @@
-export const isValidText = (text) => (text.trim()).length !== 0;
+export const isValidText = text =>
+  typeof text === 'string' && text.trim().length !== 0;
```
vaclavm: `text.trim().length !== 0` can be simplified to  `!!input`

src/utils/initialListValues.js
```diff
@@ -0,0 +1,18 @@
+import { ListItemRecord } from '../models/ListItemRecord';
```
vaclavm: Move this whole file to proper location, something like constants/initialListValues.js.

src/utils/initialListValues.js
```diff
@@ -0,0 +1,18 @@
...
+import { generateGuid } from './generateId';
+
+const initItem1 = new ListItemRecord({
```
vaclavm: Even in here I insist 👼 that clean naming should be used. makeCoffeeItem and sleepItem would be just fine.

src/components/ViewItem.jsx
```diff
@@ -21,7 +22,7 @@ export class ViewItem extends React.PureComponent {
             className="text-left"
             onClick={this._enableEditing}
           >
-            {this.props.item.index + '. ' + this.props.item.text}
+            {this.props.index + '. ' + this.props.item.text}
```
vaclavm: Rather than concatenation use this:
````
{this.props.index}.&nbsp;
{this.props.item.text}
````

File: src/utils/generateId.test.js [hunk 6]
```diff
@@ -1,13 +1,14 @@
-import { generateGuid } from '../../src/utils/generateId';
+import { generateGuid } from './generateId';
 
 describe('Guid function validation', () => {
```
vaclavm: Move the below it() to this describe() as it test same method. describe() tells you that you describe tests for specified functionality and in both of your describe() you test generateGuid ().

Also, describe() message with it() message should create a sentence that tells you what is tested and what is the result, something like: _generateGuid function generates valid format of guid_

src/utils/validateText.test.js
```diff
@@ -0,0 +1,25 @@
+import { isValidText } from './validateText';
```
vaclavm: Add empty lines between it() statements and revalidate messages as I wrote in the above (previous) comment.

src/utils/validateText.js
petrareb: !!text.trim() should work ;)

src/components/ItemToAdd.jsx
```diff
@@ -10,17 +10,22 @@ export class ItemToAdd extends React.PureComponent {
     onAddClick: PropTypes.func.isRequired
   };
 
-  state = { text: '' };
+  state = {
+    text: '',
+    enableInputColors: false
```
vaclavm: Is this property really needed? I am not convinced, but you can keep it right now and we can discuss it in person.

src/utils/validateText.js
```diff
@@ -1,2 +1,2 @@
 export const isValidText = text =>
-  typeof text === 'string' && text.trim().length !== 0;
+  typeof text === 'string' && !!text.trim();
```
vaclavm: Great! One more thing to broaden your knowledge: to keep JS untyped, you could write the check like this `text && text.trim && !!text.trim()`. In that case, you would follow JS mindset.

PR #130, opened 2018-07-26
Feature/task 3

HanaNavratilova:


src/containers/Item.jsx
```diff
@@ -0,0 +1,17 @@
+import { connect } from 'react-redux';
+import { Item as ItemComponent } from '../components/Item';
+import {
+  editItem,
+  deleteItem
+} from '../actions/ListActions';
+
+const mapStateToProps = (state, ownProps) => ({
+  item: state.list.get(ownProps.id)
```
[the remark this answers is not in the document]
PetrSvirak: It would be nice to ease the `Item` component and move its `isActive` state into the redux store as well. It will allow you to do magic (that is nothing but voluntary part of onboarding) like rejecting changes in all edited items at once or closing an item when it gets refreshed from the server… 😃

src/reducers/app.js
```diff
@@ -2,4 +2,4 @@ import { combineReducers } from 'redux';
 
 import { list } from './list';
 
-export const AppReducer = combineReducers({ list });
+export const appReducer = combineReducers({ list });
```
PetrSvirak: It should be just called `app` 😉

src/reducers/list.test.js
```diff
@@ -6,16 +6,18 @@ import { ListItem } from '../models/ListItem';
...
 import {
-  editItem,
-  deleteItem
+  saveItem,
+  deleteItem,
+  toggleItem
 } from '../actions/ListActions';
 
-const createItem = (id, text) =>
+const createItem = (id, text, isActive) =>
```
PetrSvirak: Consider providing the `isActive` with a default value - the Record's default and most often used value here seems to be `false`, so you can specify the `isActive` flag only optionally in the few `true` cases…

src/containers/Item.jsx
```diff
@@ -0,0 +1,21 @@
+import { connect } from 'react-redux';
+import { Item as ItemComponent } from '../components/Item';
+import {
+  saveItem,
+  deleteItem,
+  toggleItem
+} from '../actions/ListActions';
+
+const mapStateToProps = ({ list }, { id, index }) => ({
+  index,
```
PetrSvirak: You don't need to pass own props through `mapStateToProps` or `mapDispatchToProps` - they get added to the `props` object that `react-redux` sends to the component as `props` in the default `mergeProps` function implementation \[[docs](https://github.com/reduxjs/react-redux/blob/master/docs/api.md#arguments)].

If you were wondering why would you customize the `mergeProps` function, there is the most frequent use case [in this article](http://enthudrives.com/blog/connect-with-mergeprops/).

src/reducers/item.js
```diff
@@ -0,0 +1,22 @@
...
+  switch (action.type) {
+    case ActionType.AddItem:
+      return new ListItem({
+        id: action.payload.id,
+        text: action.payload.text,
+        isActive: false
+      });
+
+    case ActionType.ToggleItem:
+      return state.merge({ isActive: action.payload.isActive });
```
PetrSvirak: If it is a toggle, shouldn't this rather be just a negative value of current `isActive`? You would then not need payload for this action at all, would you?

src/utils/getKeys.test.js
```diff
@@ -0,0 +1,42 @@
+import { getKeys } from './getKeys';
+import { OrderedMap } from 'immutable';
+
+describe('getKeys', () => {
+  it('returns correct array', () => {
```
PetrSvirak: You somehow lost the memoization itself from this test 😆

PetrSvirak: More like `returns same [reference|array] for 2 consecutive calls with the same argument` - what do you think?

src/utils/getKeys.test.js
```diff
@@ -0,0 +1,42 @@
...
+      [854, 'dfg']
+    ]);
+    const expectedIds = [123, -489, 854];
+
+    const actualIds = getKeys(list);
+
+    expect(actualIds).toEqual(expectedIds);
+  });
+
+  it('memoization works correctly on 2 lists', () => {
```
PetrSvirak: The `works correctly` or `as expected` or … is quite problematic since people with different context have different expectations. Be more specific what test actually tests - e.g. `only memoizes consecutive
calls and does not fail when the number of arguments changes between calls`.

PR #135, opened 2018-08-09
Feature/task 3

petrareb:


src/reducers/items.js
```diff
@@ -0,0 +1,41 @@
...
+  SAVE_ITEM,
+  TOGGLE_EDITED
+} from '../constants/actionTypes';
+import { generateGuid } from '../utils/generateId';
+import { ListItemRecord } from '../models/ListItemRecord';
+import { OrderedMap } from "immutable";
+import { initialValues } from '../constants/initialListValues';
+import { item } from './item';
+
+export const newItemFactory = (generatingIdFunction) => (text) =>
```
vaclavm: One of the core principles of Redux is that reducers are pure functions. They should update the state only with values that the reducer gets from the action payload.
Don't get me wrong, it is ok to create new ListItemRecord() here in reducer and add it to state. Only the id shouldn't be generated here. It should be part of an action, you can create new file in actions folder that will create action with the correct payload.

src/actionCreators/publicActionCreator.js
```diff
@@ -0,0 +1,11 @@
+import {
```
vaclavm: Right now we are trying to refactor our product not to use publicActionCreator but to have structure like this:
````
constants/
  |-- <domain>actionTypes.ts
actions/
  |-- thunks/
      |-- someAction.ts
      |-- someOtherAction.ts
  |-- thunk<domain>Actions.ts
  |-- <domain>Actions.ts
````
Therefor you can delete this file as it has no valid usage, it only re-export actions.

File: src/actionCreators/actionCreators.js [hunk 3]
```diff
@@ -0,0 +1,35 @@
+import {
```
vaclavm: Please rename the file to be more informative - these actions apply to what (<domain>Actions.js)?

src/constants/actionTypes.js
```diff
@@ -0,0 +1,4 @@
+export const ADD_NEW_ITEM = 'ADD_NEW_ITEM';
```
vaclavm: Please rename the file so it can be distinguish for what domain they apply: <domain>actionTypes.js

src/App.jsx
```diff
@@ -1,58 +1,62 @@
...
-              </a>
-              and
-              <a href="http://redux.js.org/">
-                Redux.
-              </a>
-            </p>
-            <p>
-              You can find all the relevant info in git repository.
-            </p>
+            <div className="jumbotron">
```
vaclavm: This whole div can be deleted as you have completed the assignment ;-)

src/reducers/itemsReducer.test.js
```diff
@@ -0,0 +1,78 @@
...
+    const text = 'text';
+
+    const newItem = createNewItem(text);
+
+    expect(newItem.text).toEqual(text);
+  });
+});
+
+describe('Items reducer ', () => {
+  it('returns previous state when action is out of switch range', () => {
```
vaclavm: is out of switch range -> is unknown
You are telling that the functionality in the items reducer is switch based :)

src/reducers/itemsReducer.test.js
```diff
@@ -0,0 +1,78 @@
...
+  });
+
+  it('deletes item correctly (DELETE_ITEM action)', () => {
+    const originalStateItems = initialState.items;
+    const itemToDelete = originalStateItems.first();
+    const deletingAction = deleteItem(itemToDelete.id);
+
+    const newStateItems = items(originalStateItems, deletingAction);
+
+    expect(newStateItems.size).toEqual(originalStateItems.size - 1);
```
vaclavm: Test in this file should check that the state is changed correctly after some action and to fulfill this, you should have an expected state defined and just check that against the state after the action.
Please revalidate all other tests in this file.

src/reducers/itemsReducer.test.js
```diff
@@ -0,0 +1,78 @@
...
+  it('returns previous state when action is out of switch range', () => {
+    const invalidAction = { type: 'INVALID ACTION' };
+    const originalState = initialState;
+
+    const newState = items(originalState, invalidAction);
+
+    expect(originalState).toEqual(newState);
+  });
+
+  it('adds new item correctly (ADD_NEW_ITEM action)', () => {
```
vaclavm: After you will change how `addNewItem()` behave you can also change this test to validate against an expected value.

src/reducers/itemReducer.test.js
```diff
@@ -0,0 +1,46 @@
...
+    const itemToEdit = new ListItemRecord({
+      text: 'text',
+      id: generateGuid(),
+      isEdited: false
+    });
+    const action = toggleEdited(itemToEdit.id);
+
+    const editedItem = item(itemToEdit, action);
+
+    expect(editedItem.isEdited)
```
vaclavm: You are testing only that property of the state is changed, not the state as a whole. This test would not be able to found the problem if `item()` reducer would change something more than `isEdited`.
It is better to test for whole item, so create an expected item and validate it against the result.

src/reducers/itemsReducer.test.js
petrareb: When I was writing it, the sentence made me perfect sense... O:)

src/selectors/memoizeIds.js
```diff
@@ -0,0 +1,6 @@
+import memoize from 'memoizee';
+
+export const memoizedIds = memoize((...ids) => ids, {
```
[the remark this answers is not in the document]
petrareb: I thought the function takes just single values as a parameter, oops. I've chosen the 1st one, it looks a little bit more usable :)

src/components/App.jsx
```diff
@@ -0,0 +1,57 @@
+import '../sticky-footer.css';
+import React, { PureComponent } from 'react';
+import { Provider } from 'react-redux';
+import { TodoList } from '../containers/TodoList';
+import { store } from '../store/store';
+
+export class App extends PureComponent {
```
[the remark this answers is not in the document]
petrareb: It was done before the rebasing of master :( Sorry, I have not noticed it has changed.

src/index.jsx
```diff
@@ -1,14 +1,21 @@
...
+
 require.context('../public/', true);
 import '../node_modules/bootstrap/dist/css/bootstrap.css';
 import ReactDom from 'react-dom';
 import React from 'react';
 
 import { App } from './components/App.jsx';
+import { store } from './store/store';
 
 // ReactDom.render(
```
vaclavm: Get rid of this comment or use it in task 4 :)

PR #139, opened 2018-08-13
Feature/task 0

JarekParal:
Add Travis.yml and badge to README.md


.travis.yml
```diff
@@ -0,0 +1,6 @@
+language: node_js
+node_js:
+  - "8"
```
HanaNavratilova: It may happen that you have different version of nodejs locally (what you'd installed), in travis (in this file) and in production (in package.json file). That can cause that your solution is build locally and in travis without problems, but in production it fails. Finding the problem could be quite hard and having the same version of nodejs at least here and in package.json file could be helpful.

.travis.yml
```diff
@@ -0,0 +1,6 @@
...
+install:
+  npm install --no-optional
```
HanaNavratilova: It is definitely correct to use this command, however consider deleting this line - the default command is `npm ci` which can be significantly faster than a regular npm install (read [the description](https://docs.npmjs.com/cli/ci#description)).

.travis.yml
```diff
@@ -0,0 +1,6 @@
...
+  npm start
```
HanaNavratilova: Delete this line, please. The default [build script](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#default-build-script) is `npm test`, so you don't need run the script explicitly. Also it is written in your travis job log that `start` command is deprecated.

JarekParal:
Thanks for the review. I think that now should be everything OK - a0a42b8.

PR #140, opened 2018-08-14
Edit README.md

JarekParal:
- fix typo
## Improve Task-0 description:
- add info about a badge
- add info about eslint and tslint


PetrSvirak:
Thanks for fixing 😃 
There is a badge in the commit that should really not be there - you might want to check out a brand new branch for `upstream/master` locally…

JarekParal:
Thanks for the review. Excuse me for the badge. I checked out a new branch from the upstream/master but this bug was due to copa&paste from previous PR (#139), which was in master (hasn't own branch for this fix) and after finish task-0 and merge with branch `develop` I broke the previous PR.

Now it would be OK.

PR #143, opened 2018-08-21
Feature/task 4

petrareb:


src/utils/generateId.test.ts
```diff
@@ -1,22 +1,23 @@
...
 
     expect(result)
       .toBeTruthy();
   });
 
   it('generates different guids', () => {
-    const guid1 = generateGuid();
-    const guid2 = generateGuid();
-    const guid3 = generateGuid();
+    const guid1: Guid = generateGuid();
```
vaclavm: You don't have to specify the type to everything. Especially in tests it is a little bit [overengineering](https://vignette.wikia.nocookie.net/theslenderman/images/b/b0/Oh-boy-here-we-go.jpg/revision/latest?cb=20170526114435).

src/store/store.ts
```diff
@@ -6,12 +6,17 @@ import {
 } from 'redux';
 import logger from 'redux-logger';
 import { items } from '../reducers/items';
+import { OrderedMap } from 'immutable';
+import { initialValues } from '../constants/initialListValues';
 
-const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose;
+const composeEnhancers = (<any>window).__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose;
```
vaclavm: Simpler solution is using [composeWithDevTools from redux-devtools-extension](https://medium.com/@zalmoxis/improve-your-development-workflow-with-redux-devtools-extension-f0379227ff83).
````
const store = createStore(
  application,
  initialState,
  composeWithDevTools(),
);
````

File: src/selectors/memoizeIds.test.ts [hunk 3]
```diff
@@ -1,17 +1,18 @@
 import { memoizedIds } from './memoizeIds';
+import { Guid } from '../utils/ownTypes';
 
 describe('Function memoizeIds ', () => {
   it('returns the same values it takes as a parameter', () => {
     const originalParams = ['1', '2'];
-    const memoizedParams = memoizedIds(originalParams);
+    const memoizedParams: Array<Guid> = memoizedIds(originalParams);
```
vaclavm: Overengineering

src/utils/generateId.test.ts
petrareb: O:)

src/actions/todoActions.ts
```diff
@@ -0,0 +1,32 @@
...
+  SAVE_ITEM,
+  TOGGLE_EDITED
+} from '../constants/todoActionTypes';
+import { generateGuid } from '../utils/generateId';
+import { addNewItemActionFactory } from './factories/addNewItemActionFactory';
+import { Action } from './types/Action';
+
+export const addNewItem: (text: string) => Action = addNewItemActionFactory(generateGuid);
+
+export const toggleEdited = (id: string): Action => ({
```
vaclavm: You have defined new type for id in global.d.ts, so use it in the application everywhere you work with id.

src/containers/TodoListItem.tsx
```diff
@@ -0,0 +1,18 @@
+import { connect } from 'react-redux';
+import * as React from 'react';
+import {
+  TodoListItemStateProps,
+  TodoListItem, TodoListItemOwnProps,
+} from '../components/TodoListItem';
+import { AppState } from '../store/AppState';
+
+export interface ITodoListItemContainerOwnProps extends TodoListItemOwnProps {
```
vaclavm: TodoListItem has only one own/container props - index and id. You should create type for it in one place (component) and just import the type here and use it. No need to divide it into two separated types for component/container.

src/components/EditItem.tsx
```diff
@@ -1,50 +1,72 @@
...
+import { ListItem } from '../models/ListItem';
+import { ReactNode } from 'react';
 
-export class EditItem extends React.PureComponent {
+export type EditItemOwnProps = {
+  readonly item: ListItem,
+  readonly index: number
+};
+
+export type EditItemDispatchProps = {
```
vaclavm: You can have these without itemId as an input argument. look at comment in container.

src/components/EditItem.tsx
```diff
@@ -1,50 +1,72 @@
...
-  constructor(props) {
+  constructor(props: EditItemProps) {
     super(props);
     this.state = {
       text: props.item.text
     };
   }
 
-  _cancelEditing = () => this.props.onCancelClick(this.props.item.id);
+  _cancelEditing = (): Action => this.props.onCancelClick(this.props.item.id);
```
vaclavm: This can be simplified by calling this.props.onCancelClick in the render method. Same _deleteItem method.

src/containers/EditItem.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { EditItem, EditItemDispatchProps, EditItemOwnProps } from '../components/EditItem';
+import {
+  deleteItem,
+  saveItem,
+  toggleEdited
+} from '../actions/todoActions';
+import { Dispatch } from 'redux';
+import { Action } from '../actions/types/Action';
+
+const mapDispatchToProps = (dispatch: Dispatch): EditItemDispatchProps => ({
```
vaclavm: You can access ownProps even in dispatchToProps function and therefore get itemId directly from item passed to this component from parent.

src/components/EditItem.tsx
```diff
@@ -1,50 +1,72 @@
...
+
+export type EditItemDispatchProps = {
+  onCancelClick: (itemId: Guid) => Action,
+  onSaveClick: (itemId: Guid, text: string) => Action,
+  onDeleteClick: (itemId: Guid) => Action
+};
+
+export type EditItemProps = EditItemDispatchProps & EditItemOwnProps;
+
+export type EditItemState = {
```
vaclavm: You do not need to export this type.

src/components/TodoListItem.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export type TodoListItemOwnProps = {
+  readonly index: number
+};
+
+export type TodoListItemStateProps = {
+  readonly item: ListItem
+};
+
+export type TodoListItemProps = TodoListItemStateProps & ITodoListItemContainerOwnProps;
```
vaclavm: You do not need to export this type

src/containers/ViewItem.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { connect } from 'react-redux';
+import {
+  ViewItemDispatchProps,
+  ViewItem,
+  ViewItemOwnProps
+} from '../components/ViewItem';
+import { toggleEdited } from '../actions/todoActions';
+import { Dispatch } from 'redux';
+
+const mapDispatchToProps = (dispatch: Dispatch): ViewItemDispatchProps => ({
```
vaclavm: Use ownProps to access item.id instead of passing one as an argument from component.

src/components/ViewItem.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { Action } from '../actions/types/Action';
+import { ReactNode } from 'react';
+
+export type ViewItemOwnProps = {
+  readonly index: number,
+  readonly item: ListItem
+};
+
+export type ViewItemDispatchProps = {
+  onRowClick: (id: Guid) => Action
```
vaclavm: Should be readOnly and you do not need the id here, you get it from ownProps in container.

src/components/ViewItem.tsx
```diff
@@ -0,0 +1,43 @@
...
+};
+
+export type ViewItemProps = ViewItemDispatchProps & ViewItemOwnProps;
+
+export class ViewItem extends React.PureComponent<ViewItemProps> {
```
vaclavm: The whole component can be made functional once you update the container, therefore _enableEditing can be removed.

src/models/baseItem.ts
```diff
@@ -0,0 +1,14 @@
+import { Record } from 'immutable';
+
+export function baseItem<T>(defaultValues: T, name: string): any {
+  class BaseItem extends Record(defaultValues, name) {
+    constructor(props: Partial<T>) {
+      super(props);
+    }
+
+    with = (item: Partial<T>) =>
```
vaclavm: Well, for baseItem it is not item. It should be generic entity, property, ...

src/models/baseItem.ts
```diff
@@ -0,0 +1,14 @@
+import { Record } from 'immutable';
+
+export function baseItem<T>(defaultValues: T, name: string): any {
```
vaclavm: It should have more generic name, like baseRecord ... baseItem tells you, it is a base for Item, but what if you would add person model?

src/components/EditItem.tsx
vaclavm: Also, I would consider naming these a little bit differently. Right now you read these in the component like this (take delete button for instance): onClick do onDeleteClick. Would it be better to look at component and read it like this (again delete button): onClick do deleteItem?

src/containers/TodoListItem.tsx
petrareb: Index and id were divided because the index is not used by this component, but by his successor ViewItem/EditItem. (ask Peťo) :)

src/containers/ViewItem.tsx
```diff
@@ -0,0 +1,16 @@
...
+import {
+  ViewItemDispatchProps,
+  ViewItem,
+  ViewItemOwnProps
+} from '../components/ViewItem';
+import { toggleEdited } from '../actions/todoActions';
+import { Dispatch } from 'redux';
+
+const mapDispatchToProps = (dispatch: Dispatch, ownProps: ViewItemOwnProps): ViewItemDispatchProps => ({
+  onRowClick: () => dispatch(toggleEdited(ownProps.item.id))
```
vaclavm: Well, what if there would be no rows in your solution? What would happen? 👼

src/components/ItemToAdd.tsx
```diff
@@ -0,0 +1,82 @@
+import * as React from 'react';
+import classNames from 'classnames';
+import * as PropTypes from 'prop-types';
+import { isValidText } from '../utils/validateText';
+import { ReactNode } from 'react';
+
+export type ItemToAddDispatchProps = {
+  readonly addItem: (text: string) => void
```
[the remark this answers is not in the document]
petrareb: Oops O:) I forgot it there and didn't notice while checking

PR #147, opened 2018-08-22
Feature/task 2

VeronikaLou:


src/components/List.jsx
```diff
@@ -15,49 +13,55 @@ export class List extends PureComponent {
...
+    const items = this.state.items
+      .set(newItem.id, newItem);
+    this.setState(() => ({ items }));
   };
 
   _saveChanges = (id, text) => {
-    const items = this.state.items.map(item => ((item.id === id)
-      ? createItem(id, text)
-      : item));
+    const item = {
```
vaclavm: This whole method can be changed to one line (also other methods in this file can be refactored):
````
  _saveChanges = (id, text) =>
    this.setState((prevState) => ({
      items: prevState.items
        .mergeIn([id], {
          text,
          isEdited: false
        })
    }));
````

Comment (vaclavm) [reply src/components/List.jsx]: Also, it is recommended to use prevState in functional setState, see React docs:
> React may batch multiple setState() calls into a single update for performance. Because this.props and this.state may be updated asynchronously, you should not rely on their values for calculating the next state. To fix it, use a second form of setState() that accepts a function rather than an object. That function will receive the previous state as the first argument, and the props at the time the update is applied as the second argument.

File: src/models/ListItem.jsx [hunk 2]
```diff
@@ -0,0 +1,9 @@
+import { Record } from 'immutable';
+
+const emptyItem = {
+  id: '00000000-0000-0000-0000-000000000000',
+  text: '',
+  isEdited: false
+};
+
+export const ListItem = new Record(emptyItem);
```
vaclavm: Record can be named by the second parameter, it is better for debugging the application.

PR #149, opened 2018-08-27
Feature/task 1

RadekP3:


src/components/List.jsx
```diff
@@ -1,34 +1,99 @@
...
+      alert('You have to enter some text!');
+      return;
+    }
+    const newItem = {
+      id: generateId(),
+      value: itemText
+    };
+
+    this.setState(prevState => ({
+      items: {
```
kubmir: Here you change the type of items from an array to object. It is the reason why you handle operations with array quite weird. Items should be stored in array so you need to make something like this
```
items: [
    ...previousState.notes,
    noteToAdd,
],
```
This will make a new array with the added item - your code make a new object where items are stored as properties.

Also consider using map and filter functions when working with this array.

PR #155, opened 2018-09-04
Feature/task 2

RadekP3:


src/components/List.jsx
```diff
@@ -1,64 +1,59 @@
...
-        ...item,
-        isInEditMode: false
-      }))
+      items: prevState.items.setIn([itemId, 'isInEditMode'], false),
     }));
   };
 
   _renderListItems = () =>
-    this.state.items.map((item, index) => (
+    this.state.items.valueSeq().map((item, index) => (
```
kubmir: You can improve readability of the code with formatting like this one:
```
this.state.notes
      .valueSeq()
      .map((note, i) => {})
```

src/models/ListItem.jsx
```diff
@@ -0,0 +1,9 @@
+import { Record } from 'immutable';
+
+const defaultItem = {
```
kubmir: You named Record as a ListItem. This object describes the default values of the ListItem object. I suggest renaming to the defaultListItem.

src/components/List.jsx
```diff
@@ -1,64 +1,59 @@
...
     }));
   };
 
   _clickLabel = (itemId) => {
     this.setState(prevState => ({
-      items: prevState.items.map(item => (item.id !== itemId ? item : {
-        ...item,
-        isInEditMode: true
-      }))
+      items: prevState.items.setIn([itemId, 'isInEditMode'], true),
```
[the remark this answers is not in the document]
RadekP3: Is it ok if I use mergeIn here? Merge doesn't seem to work for me.

src/models/ListItem.js
```diff
@@ -6,4 +6,4 @@ const defaultListItem = {
   isInEditMode: false
 };
 
-export const ListItem = new Record(defaultListItem);
+export const ListItem = new Record(defaultListItem, 'list item');
```
vaclavm: The naming convention is to name the record same as the const/file. That would be ListItem. You can fix this into task 3.

PR #157, opened 2018-09-05
Feature/task 4

VeronikaLou:


src/actions/IListAction.ts
```diff
@@ -0,0 +1,4 @@
+export interface IListAction {
```
vaclavm: Move this to inner folder types

src/components/EditedItem.tsx
```diff
@@ -1,9 +1,30 @@
...
+
+export interface IEditedItemStateProps {
+  readonly text: string;
+}
+
+interface IEditedItemState {
+  readonly text: string;
+}
+
+type EditedItemProps = IEditedItemOwnProps & IEditedItemDispatchProps & IEditedItemStateProps;
```
vaclavm: You are mixing interfaces with type aliases for possible type intersections. Please be consistent and use only interfaces/types for your components and containers.
`interface IEditedItemProps extends IEditedItemOwnProps, IEditedItemDispatchProps, IEditedItemStateProps { }`

Check this in all your components/containers.

src/models/IStore.ts
```diff
@@ -0,0 +1,6 @@
+import { OrderedMap } from 'immutable';
+import { ListItem } from './ListItem';
+
+export interface IStore {
+  items: OrderedMap<Uuid, ListItem>;
```
vaclavm: Create new file with type for your items (`OrderedMap<Uuid, ListItem>`) because you use OrderedMap<Uuid, ListItem> multiple times in your application (reducer items.ts).

src/models/IStore.ts
```diff
@@ -0,0 +1,6 @@
+import { OrderedMap } from 'immutable';
+import { ListItem } from './ListItem';
+
+export interface IStore {
```
vaclavm: Create new folder store and move there this file and the newly creater ItemsState.ts file ... move the to inner folder types inside newly create store folder.

src/utils/store.ts
```diff
@@ -1,22 +1,25 @@
 import {
```
vaclavm: Move this file to store folder (mentioned previously at IStore.ts)

src/utils/getMemoizedIds.ts
```diff
@@ -0,0 +1,7 @@
+import * as memoizee from 'memoizee';
+import { Seq } from 'immutable';
```
vaclavm: I don't like that you are referencing to immutable here ... I think it would be better to pass array here instead of keySeq(). So that the getMemoizedIds would be independent on implementation.

src/models/ListItem.tsx
```diff
@@ -0,0 +1,27 @@
...
+  readonly isEdited: boolean;
+}
+
+const emptyItem: IListItem = {
+  id: '00000000-0000-0000-0000-000000000000',
+  text: '',
+  isEdited: false,
+};
+
+export class ListItem extends Record(emptyItem, 'ListItem') {
```
vaclavm: I have totally forgotten this ListItem! 
BTW good job, the item looks neat but we need to take it a little bit further.

Imagine another model that would have different properties but should also have constructor and with method. You would have to again define constructor and with method. How about having one generic base class that would extend the Record and you would use your class here instead of Record. I know it is tough to imagine, so don't hesitate to ask anyone about base class for your record.

PR #159, opened 2018-09-11
Feature/task 4

JarekParal:


src/components/ShowItem.tsx
```diff
@@ -0,0 +1,31 @@
...
+  readonly onEditStart: () => void;
+}
+
+export interface IShowItemStateProps {
+  text: string;
+}
+
+export interface IShowItem extends IShowItemProps, IShowItemStateProps { }
+
+export const ShowItem: React.SFC<IShowItem> = (props) => (
```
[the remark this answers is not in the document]
JarekParal: I was thinking that `React.SFC<IShowItem>` define types for the `props`. Isn't that enough?  

2018-09-11 15_57_41-greenshot

src/components/EditItem.tsx
```diff
@@ -1,23 +1,39 @@
-import React, { PureComponent } from 'react';
-import PropTypes from 'prop-types';
+import * as React from 'react';
+import { PureComponent } from 'react';
+import * as PropTypes from 'prop-types';
 
-export class EditItem extends PureComponent {
+export interface IEditItemProps {
```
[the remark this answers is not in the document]
JarekParal: Do you want some change or is that just a statement?

JarekParal:
Finish

src/actions/actionCreators.ts
```diff
@@ -1,23 +1,24 @@
...
   EDIT_ITEM,
-  DELETE_ITEM
+  DELETE_ITEM,
 } from './actionTypes';
 import { createItemFactory } from './actionCreatorsFactory';
 
 export const createItem = createItemFactory(generateId);
 
-export const editItem = (id, text) => ({
+export const editItem = (id: Guid, text: string): IAction => ({
```
PetrSvirak: I'm not sure whether _editItem_ is too generic. What you actually update is item's text. This name could mislead you: if you decide to edit some other aspect of the item, you might be tempted to extend this action rather than creating a new one, describing user's action.

src/actions/actionCreators.ts
```diff
@@ -1,23 +1,24 @@
+import { IAction } from './IAction';
 import { generateId } from '../utils/generateId';
 import {
   EDIT_ITEM,
-  DELETE_ITEM
+  DELETE_ITEM,
 } from './actionTypes';
 import { createItemFactory } from './actionCreatorsFactory';
 
 export const createItem = createItemFactory(generateId);
```
PetrSvirak: It would be better to separate dependecy injection from (basic) action definitions. You create an `index.ts` in this folder end perform DI there - just move `createItem` there and re-export (`export from`) other action creators from this files. Alternatively, you could go the Cloud way, however, DI is not used in simple actions there and thus they don't have this case covered.

JarekParal: What is the 'Cloud way'? Could you give me please some link?

src/components/EditItem.tsx
```diff
@@ -1,11 +1,29 @@
...
+export interface IEditItem extends IEditItemProps, IEditItemStateProps, IEditItemDispatchProps {}
+
+export class EditItem extends PureComponent<IEditItem, IEditItemStateProps> {
   static displayName = 'EditItem';
 
   static propTypes = {
-    text: PropTypes.string.isRequired,
+    id: PropTypes.string.isRequired,
     position: PropTypes.number.isRequired,
     finishEdit: PropTypes.func.isRequired,
```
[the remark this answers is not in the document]
JarekParal: Is it `onFinishEdit` suitable?

src/components/AddItem.tsx
```diff
@@ -1,23 +1,37 @@
-import React, { PureComponent } from 'react';
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import { PureComponent } from 'react';
 
-export class AddItem extends PureComponent {
+export interface IAddItemProps {
+  readonly onChange: (input: string) => void;
```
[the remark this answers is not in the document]
JarekParal: Is it the correct name for the event `onClick`?

src/components/Item.tsx
```diff
@@ -0,0 +1,57 @@
...
+export class Item extends PureComponent<IItemProps, IItemState> {
+  static displayName = 'Item';
+
+  static propTypes = {
+    id: PropTypes.string.isRequired,
+    position: PropTypes.number.isRequired,
+  };
+
+  state = {
+    isEdited: false,
```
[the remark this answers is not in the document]
JarekParal: Should I add it to `store`? @PetrSvirak

src/components/AddItem.tsx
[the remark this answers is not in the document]
PetrSvirak: It's one many I would see fit.

src/components/Item.tsx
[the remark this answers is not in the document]
PetrSvirak: Preferably yes 😃

src/components/EditItem.tsx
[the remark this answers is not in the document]
PetrSvirak: Not really, I would you like to try to match the name with other button-related callbacks.

src/actions/actionCreators.ts
PetrSvirak: They separate actions by "origin", so they have 2 files - "actionCreators" (for basic actions (objects)) and "thunkActionCreators" (for thunks used usually for asynchronous tasks triggered by a user action). In this case, you would have to mend together pure uninjected creators with the injected ones and that is something I believe to combine 2 responsibilities and result in an unnecessarily long file(s). I have to add that it makes sense in Cloud since they a) don't use dependency injection in basic actions and b) rely heavily on thunks. Sorry, I should describe it in clear way.

src/components/ShowItem.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+
+export interface IShowItemProps {
+  readonly id: string;
```
JarekParal: Should I have a type for `id` rather my type [`Guid`](https://github.com/KenticoAcademy/kentico-onboarding-js/pull/159/files/2e3a1ba1aea065302a056d5acea67039b4c195b1#diff-0ce2f670b468151d11024c9ab2488825R1) then `string`?  @PetrSvirak

src/containers/EditItem.tsx
```diff
@@ -0,0 +1,28 @@
...
+  editItem,
+} from '../actions/actionCreators';
+import {
+  EditItem as EditItemComponent,
+  IEditItemProps,
+  IEditItemDispatchProps,
+  IEditItemStateProps,
+} from '../components/EditItem';
+
+const mapStateToProps = (state: IState, {id}: { id: GUID }): IEditItemStateProps => ({
```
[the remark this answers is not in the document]
JarekParal: Is it better to define `IOwnProps` or I can use `IEditItemOwnProps` in `components/EditItem` => `{id}: IEditItemOwnProps`?

src/components/Item.tsx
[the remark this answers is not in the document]
JarekParal: Fix 7f942cffbacfba435d29ff5ad464c446f69fd571

src/components/Item.tsx
```diff
@@ -0,0 +1,57 @@
...
+          position={this.props.position}
+          finishEdit={this._finishEditItem}
+        />)
+      : (
+        <ShowItem
+          position={this.props.position}
+          id={this.props.id}
+          onEditStart={this._startEditItem}
+        />)
+  );
```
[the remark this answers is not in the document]
JarekParal: Are you sure that it is more readable? You need to check two more functions when you want to see the code (5ac7781e100ad55ec09607ef3ada92ed27b8de53). I understand that the name of functions added new information for a programmer but added more complexity when the programmer wanted to read the code. You can also add just comments for a better description.
@PetrSvirak

src/actions/actionCreators.ts
JarekParal: Ok, that is true. Is it better name for the action `editTextItem`? @PetrSvirak

JarekParal: `saveTextItem`?

src/components/ShowItem.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export interface IShowItemStateProps {
+  readonly text: string;
+}
+
+export interface IShowItemDispatchStateProps {
+  readonly onEditStart: () => void;
+}
+
+export interface IShowItem extends IShowItemOwnProps, IShowItemStateProps, IShowItemDispatchStateProps { }
```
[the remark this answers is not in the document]
PetrSvirak: Why do you export the type? (you would likely need to export it only when defining your own `mergeProps` in the container, however, you do not do that nor import the interface anywhere 😉)

src/components/Board.tsx
```diff
@@ -0,0 +1,39 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import { AddItem } from '../containers/AddItem';
+import { Item } from '../containers/Item';
```
PetrSvirak: Consider creating a `board` folder inside your `components` and `containers` folder or create an `apps` folder and place all `reducers`, `components`, `actions`, `containers`, `models` that are related to your board under a `board` subfolder of the `apps` - that is definitely task for another time though… It just feels it would be a bit more natural to see your solution more structured… Feel free to make a note for yourself in next task Jira issue or create a Jira issue on your own (don't forget to put it in your epic 😉)

src/reducers/app.ts
```diff
@@ -0,0 +1,6 @@
+import { combineReducers } from 'redux';
+import { board } from './board/board';
+
+export const app = combineReducers({
```
PetrSvirak: You are missing type here - it literally is `any` right now 😉 (this can cause your `app` reducer not to return an object of `IState` type - even if there was the `IState` mentioned (which regrettably isn't).

src/utils/getInitialItems.ts
```diff
@@ -5,15 +5,15 @@ import { generateId } from './generateId';
...
-    text: 'Cat'
+    text: 'Cat',
   });
   const elephant = new Item({
     id: generateId(),
-    text: 'Elephant'
+    text: 'Elephant',
   });
 
   return OrderedMap({
```
PetrSvirak: What I overlooked before (sorry) is that you create the order map in the same function as initial items. Not that it is too bad, but initial items are actually just a bunch of items - an array, let's say. How you structure them into your initial state seems to be a bit different story to me (or at least a story for a different method).

src/App.tsx
```diff
@@ -1,6 +1,6 @@
 import './sticky-footer.css';
-import React from 'react';
-import { Board } from './components/Board';
+import * as React from 'react';
+import { Board } from './containers/Board';
```
PetrSvirak: You left some unused files behind when you removed `List` component from this file 😉

src/reducers/app.ts
JarekParal:
```suggestion
export const app = combineReducers({
```

JarekParal: Yes, that is true. I forgot types here. Thanks @PetrSvirak.

Should I add the type to all combinedReducers (e.g. `board`) or is it enough to set in the top level reducer and then it will be checked on all levels?

src/components/ShowItem.tsx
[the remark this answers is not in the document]
JarekParal: Why is the `type` better? Ok, it could be up to 11 characters shorter but that is not consistent with other interfaces (all other interfaces start with `interface` but this start as `type`).
You need to use different syntax to the same purpose and that can be misleading for other readers.

src/components/Board.tsx
JarekParal: https://kentico.atlassian.net/browse/KA-404

JarekParal:
Finish or ready for discussion :-)

src/components/ShowItem.tsx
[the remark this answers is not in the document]
PetrSvirak: It says the same with less words - that's one thing. The other is that you are not (that much) tempted to add any properties into the interface. I agree that it's inconsistent though. 

The main reason for me personally would be that while an interface expresses a logical unit – describes a subset of properties forming props in this case – the props interface itself is exactly an alias of the unification of the partial -props interfaces. I see that as the most distinguishing factor between the interfaces and the type; but that's just me…

src/reducers/app.ts
PetrSvirak: If you only set type in the top-most one, the `combineReducer`s below would still return `any` – thus preventing the top-most one perform any type-check…

src/reducers/board/board.ts
```diff
@@ -1,6 +1,9 @@
-import { combineReducers } from 'redux';
+import { combineReducers, Reducer } from 'redux';
 import { items } from './items/items';
+import { OrderedMap } from 'immutable';
+import { Item } from '../../models/Item';
+import { IAction } from '../../actions/IAction';
 
-export const board = combineReducers({
+export const board: Reducer<{ items: OrderedMap<Guid, Item> }, IAction> = combineReducers({
```
PetrSvirak: That's not very nice. Imagine how would the type look when you added more properties to the `board` reducer. Also, you repeat yourself - you had to say the same in the `IState` at least 😉

src/utils/getOrderMapFromItems.ts
```diff
@@ -0,0 +1,6 @@
+import { Item } from '../models/Item';
+import { OrderedMap } from 'immutable';
+
+export const getOrderMapFromItems = (items: Item[]): OrderedMap<Guid, Item> =>
+  items.reduce((orderMap, item) => { return orderMap.set(item.id, item); },
```
PetrSvirak: Nice use of the `reduce` 😃 

Why the `return` again? 👼 

Also, you create `n` new maps here since `OrderedMap` is immutable. Could you come up with a less "mappy" solution (I mean, a solution where would fewer objects be created in the process) or argue (why) it cannot be done? 😃

src/utils/getOrderMapFromItems.ts
```diff
@@ -0,0 +1,6 @@
+import { Item } from '../models/Item';
+import { OrderedMap } from 'immutable';
+
+export const getOrderMapFromItems = (items: Item[]): OrderedMap<Guid, Item> =>
```
PetrSvirak: I would be OK with this function residing in the `store.ts` file.

src/App.tsx
```diff
@@ -1,4 +1,3 @@
-import './sticky-footer.css';
```
PetrSvirak: Haven't this break your footer ('s position)? 😛 
I was mainly speaking about the `assignement.gif` 😉

JarekParal: Yes, it breaks the footer but after the restart the NPN.

src/components/ShowItem.tsx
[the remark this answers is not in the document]
JarekParal: That makes sense.

JarekParal:
Done.

src/reducers/board/items/items.ts
```diff
@@ -0,0 +1,31 @@
...
+  CREATE_ITEM,
+  SAVE_TEXT_ITEM,
+  START_EDIT_ITEM,
+  CANCEL_EDIT_ITEM,
+  DELETE_ITEM,
+} from '../../../actions/actionTypes';
+import { item } from './item';
+import { Item } from '../../../models/Item';
+
+export type IItemState = OrderedMap<Guid, Item>;
```
PetrSvirak: This one is a bit too much. Only type the `node` reducers, the `leave` reducers usually don't need special type (I mean a type that would only be created to be used within the reducers' structure)…

JarekParal: Ok @PetrSvirak, but when am I using this type in more places?

https://github.com/KenticoAcademy/kentico-onboarding-js/pull/159/commits/e19949eebaf9779f0255f6786a863029dc4ca810#diff-4d56072f9e01ddfbbafb58936cd8e4d7L5

https://github.com/KenticoAcademy/kentico-onboarding-js/pull/159/commits/e19949eebaf9779f0255f6786a863029dc4ca810#diff-314db12f3a037090af2c7a1b0d7645d1L11

JarekParal: Ok @PetrSvirak, but when am I using this type in more places?

https://github.com/KenticoAcademy/kentico-onboarding-js/pull/159/commits/e19949eebaf9779f0255f6786a863029dc4ca810#diff-4d56072f9e01ddfbbafb58936cd8e4d7L5

https://github.com/KenticoAcademy/kentico-onboarding-js/pull/159/commits/e19949eebaf9779f0255f6786a863029dc4ca810#diff-314db12f3a037090af2c7a1b0d7645d1L11

JarekParal: Ok @PetrSvirak, but when am I using this type in more places?

https://github.com/KenticoAcademy/kentico-onboarding-js/pull/159/commits/e19949eebaf9779f0255f6786a863029dc4ca810#diff-4d56072f9e01ddfbbafb58936cd8e4d7L5

https://github.com/KenticoAcademy/kentico-onboarding-js/pull/159/commits/e19949eebaf9779f0255f6786a863029dc4ca810#diff-314db12f3a037090af2c7a1b0d7645d1L11

PetrSvirak: I guess it was due to GitHub issues, but still, sending comment once is enough even with my response times 🤣 

Anyway, to your question: This is very similar to declaring type alias for a Dictionary in C# - you can declare something like `class MyDictionary : Dictionary<Guid, Item> { }`, however, it is usually not that useful since it hides what you are expected to provide or what you get (being it a method parameter or result).
My point here is that the fact that you have a dictionary consisting of 2 types is more relevant than a custom name that will have roughly similar meaning; the amount of repetition does not seem to be that high and accuracy and impossibility of altering the type and ownership of individual methods over some unifying concept that might be hard/unnecessary to grasp and bunch other things overweight the benefit of custom naming in my eyes – but if you strongly insist, leave the alias there 😃

PR #161, opened 2018-10-10
Feature/task 5

pthutta:


src/actions/creators/asyncCreators.test.ts
```diff
@@ -0,0 +1,206 @@
...
+
+    it('when fails, creates displayError action', () => {
+      const tempId: Guid = '1';
+      const text: string = 'Item text.';
+      const error: string = 'There was an error while creating new item: ' + GATEWAY_TIMEOUT_MESSAGE;
+      const expectedResult: AddItemErrorAction = addItemError(tempId, error, 'POST');
+      fetchResponse = mockResponse(504, error);
+      dispatch = jest.fn(() => addItemCreator(() => tempId)(text));
+
+      postItemCreator(fetch, addItemCreator(() => tempId), addItemError)(text)(dispatch, () => undefined, undefined)
```
medvedovic: This is not readable at all. Arrange, act, assert 😉

src/actions/creators/asyncCreators.test.ts
```diff
@@ -0,0 +1,206 @@
...
+  } from './getItemsCreator';
+import Mock = jest.Mock;
+import {
+  deleteItemSuccess,
+  displayError,
+  loadingItemsSuccess, postItemSuccess, putItemSuccess,
+  saveItemText, setItemSyncing,
+} from '../itemsActions';
+import { addItemCreator } from './addItemCreator';
+import { urlBase } from '../utils/urlBase';
```
medvedovic: You import quite a lot of things.. I don't see any reason for you to import urlBase or headersBase. These also should not test the arguments for fetch, that should be tested elsewhere. What you want to do is to make sure the dispatch was called with the right actions, that's all
Also, you may want to rewrite your tests with async/await, it looks sexier that way 😏

src/actions/creators/postItemCreator.ts
```diff
@@ -0,0 +1,36 @@
...
+  AddItemAction,
+  AddItemErrorAction,
+  PostItemSuccessAction,
+} from '../types/itemsActionTypes';
+import { Dispatch } from 'redux';
+import { urlBase } from '../utils/urlBase';
+import { postItemSuccess } from '../itemsActions';
+import { fetchFactory } from '../utils/fetchFactory';
+
+export const postItemCreator = (
```
medvedovic: Why don't you try to rewrite your thunks to async/await? 😈

medvedovic: Also, this is a function, not a class.. creator is not correct here, createPostItem would be better, createPostItemThunk even better

src/components/ListItem.tsx
```diff
@@ -19,13 +21,29 @@ const listItemPropTypes: ValidationMap<ListItemProps> = {
...
   isBeingEdited: PropTypes.bool.isRequired,
+  isSyncing: PropTypes.bool.isRequired,
+  errorId: PropTypes.string.isRequired,
 };
 
-export const ListItem: React.StatelessComponent<ListItemProps> = ({ order, id, isBeingEdited }): JSX.Element => (
-  isBeingEdited
+export const ListItem: React.StatelessComponent<ListItemProps> = ({ order, id, isBeingEdited, isSyncing, errorId}): JSX.Element => {
+  const item: JSX.Element =
+    isBeingEdited && errorId === '' && !isSyncing
```
medvedovic: Nope. The component should not consider itself with the properties of the particular item. That way you're breaking the reusability of the component. I'd suggest you return one flag from the container only

src/containers/DisplayListItem.tsx
```diff
@@ -12,10 +12,15 @@ import { IAppState } from '../store/state/IAppState';
 
 const mapStateToProps = (state: IAppState, ownProps: DisplayListItemContainerProps): DisplayListItemStateProps => ({
   text: state.list.items.get(ownProps.id).text,
+  hasError: state.list.items.get(ownProps.id).errorId !== '',
 });
 
 const mapDispatchToProps = (dispatch: Dispatch, ownProps: DisplayListItemContainerProps): DisplayListItemDispatchProps => ({
-  enableEditing: () => dispatch(toggleItemEditing(ownProps.id)),
+  enableEditing: () => {
+    if (!ownProps.isSyncing) {
```
medvedovic: Nope, this is clutter. The decision should be made by the component not the container

src/@types/global.d.ts
```diff
@@ -1,5 +1,11 @@
+import { ThunkAction } from 'redux-thunk';
+import { Actions } from '../actions/types/itemsActionTypes';
+
 export {};
 
 declare global {
   type Guid = string;
+  type ErrorAction = 'POST' | 'PUT' | 'DELETE';
```
medvedovic: These are not actions 😉 Add, Update, Delete and so forth are..

src/actions/thunkFactories/deleteItemFactory.ts
```diff
@@ -0,0 +1,32 @@
...
+  readonly generateErrorId: () => Guid;
+}
+
+export const deleteItemFactory = (dependencies: IDeleteItemDependencies) =>
+  (id: Guid): Thunk =>
+    async (dispatch: Dispatch): Promise<DeleteItemSuccessAction | AddItemErrorAction> => {
+      dispatch(setItemSyncing(id, true));
+
+      try {
+        const response: Response = await dependencies.fetch(`/${id}`, {method: 'DELETE'});
```
medvedovic: This one applies to all thunks. It should not decide about the location or method to use.. Considering postItemFactory, it should not do anything with the data to send
In other words, the injected fetch should be a function, that takes an input (data to send) and do all the heavy lifting out of the box, it is not a job for the thunk

src/models/Item.ts
```diff
@@ -3,18 +3,27 @@ import { ModelBase } from './ModelBase';
 export interface IItem {
   readonly id: Guid;
   readonly text: string;
+  readonly oldText: string;
```
medvedovic: This does not belong here... I like that you tried to implement some kind of fallback policy, though I don't like the realization.. Try to come up with a better approach

src/actions/thunkActions/cancelThunkAction.ts
```diff
@@ -0,0 +1,31 @@
...
+import { IError } from '../../models/Error';
+import { Actions } from '../types/itemsActionTypes';
+import {
+  cancelItemUpdating,
+  deleteItemError,
+  deleteItemSuccess,
+} from '../itemsActions';
+
+export const cancelThunkAction = (itemId: Guid) =>
+  (dispatch: Dispatch, getState: () => IAppState): Actions | undefined => {
```
medvedovic: Change return type to `Actions | void`

src/actions/thunkFactories/thunkFactories.test.ts
```diff
@@ -0,0 +1,281 @@
...
+import { putItemFactory } from './putItemFactory';
+import { deleteItemFactory } from './deleteItemFactory';
+import { GATEWAY_TIMEOUT_MESSAGE } from '../utils/errorMessages';
+import Mock = jest.Mock;
+import { ErrorActionType } from '../types/ErrorActionType';
+import { ItemError } from '../../models/Error';
+import { OrderedMap } from 'immutable';
+
+describe('fetchCreators', () => {
+  const mockResponse = (status: number, statusText?: string, body?: BodyInit) =>
```
medvedovic: This is not necessary, actually, it's redundant.. You wrote down a lot of logic for creating a simple promise containing response object. In other words, you mock a response object to mock `fetch` functionality. Why not mocking fetch from the git-go?
Something like
```
const mockFetch = (isOk: boolean, data: any) =>
  () =>
    isOk
      ? Promise.resolve(new Response(data))
      : Promise.reject(new Response());
```
is quite enough

medvedovic: Also, this file is a little too long 😉

src/actions/thunkFactories/thunkFactories.test.ts
```diff
@@ -0,0 +1,281 @@
...
+
+  describe('getItemsFactory', () => {
+    it('when successful, creates loadingItemsSuccess action', async () => {
+      // arrange
+
+      const fetchedItems: IItem[] = [
+        new Item({id: '1', text: 'Text1'}),
+      ];
+      const expectedResult: LoadingItemsSuccessAction = loadingItemsSuccess(fetchedItems);
+      response = mockResponse(200, undefined, '[{"id":"1","text":"Text1"}]');
```
medvedovic: If I were you, I'd let the mocking function stringify the object rather than passing a "hardcoded json"

src/actions/thunkFactories/retryThunkActionFactory.test.ts
```diff
@@ -0,0 +1,151 @@
...
+              id: errorId,
+              message: 'Error2',
+              action: ErrorActionType.Delete,
+            }),
+          ],
+        ]),
+      },
+    }));
+
+    await retryThunkActionFactory({ deleteItemRequest, repostItemRequest, putItemRequest })(id)(dispatch, getState, undefined);
```
medvedovic: Didn't I ask you to split these? 😛

src/@types/global.d.ts
```diff
@@ -1,5 +1,12 @@
...
+import { ThunkAction } from 'redux-thunk';
+import { Actions } from '../actions/types/itemsActionTypes';
+import { ErrorActionType } from '../actions/types/ErrorActionType';
+import { IAppState } from '../store/state/IAppState';
 
 declare global {
   type Guid = string;
+  type ErrorAction = ErrorActionType.Add | ErrorActionType.Update | ErrorActionType.Delete;
+  type Fetch = (input: string, init: RequestInit) => Promise<Response>;
+  type ItemFetch = (id?: Guid, data?: BodyInit) => Promise<Response>;
```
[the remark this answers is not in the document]
medvedovic: Talking about "omnipotent code", I'd like to share [an article](https://web-techno.net/dry-principle-explained/) about DRY principle with you... 
**TL;DR**
There is a big misconception regarding the principle, or at least, there used to be with me. DRY does not mean that if you have "two pieces of code" that look alike you should find a more generic solution, as to stop "duplicating" code. It means that "knowledge" or "business logic" in your app should not be duplicated... What could be considered "knowledge"? For example, a class that parses B-trees (example taken from the article).. You wouldn't implement this algorithm in 50 places in your codebase. Hope this helps

src/components/EditListItem.tsx
```diff
@@ -47,62 +48,79 @@ export class EditListItem extends PureComponent<EditListItemProps, EditListItemS
...
+        <form className="flexbox" onKeyPress={this._onKeyPress}>
+          <section className={className}>
+            <div className="stretch">
               <input
+                type="text"
                 className="form-control"
                 value={this.state.inputText}
                 onChange={this._storeInputValue}
+                tabIndex={this.props.order}
+                autoFocus={true}
```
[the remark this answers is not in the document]
pthutta: Current configuration of TSLint demands setting of values of boolean attributes.

src/actions/thunkActions/itemRequests.ts
```diff
@@ -0,0 +1,54 @@
...
+    id,
+    data: JSON.stringify({id, text}),
+  });
+};
+
+export const deleteItemRequest = async (id: Guid): Promise<IItem> => {
+  const deleteItem = itemFetch({ method: 'DELETE' });
+  const response = await deleteItem({ id });
+
+  const data = await response.json();
```
PetrSvirak:
```suggestion
  const serverItem = await response.json();
```
-- or --
```suggestion
  const rawItem = await response.json();
```
-- or --
…
😛

src/actions/thunkActions/itemRequests.ts
```diff
@@ -0,0 +1,54 @@
+import {
+  IItem,
+  Item,
+} from '../../models/Item';
+import { fetchFactory } from '../thunkFactories/fetchFactory';
+import { urlBase } from '../utils/urlBase';
+import { headerBase } from '../utils/headerBase';
+
+const itemFetch = fetchFactory(fetch, urlBase, headerBase);
```
PetrSvirak: This file (`itemRequests.ts`) along with the `fetchFactory` actually belongs to the `utils` folder, doesn't it? I mean, none of these are thunks, both can be used out of thunk actions if needed be…

src/actions/thunkFactories/putItemFactory.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export interface IPutItemDependencies {
+  readonly putItemRequest: (item: IItem) => Promise<void>;
+  readonly createItemError: (message: string, errorType: ErrorAction, itemId: Guid) => AddItemErrorAction;
+}
+
+export const putItemFactory = (dependencies: IPutItemDependencies) =>
+  (id: Guid, text: string, oldText?: string): Thunk =>
+    async (dispatch: Dispatch, getState: () => IAppState): Promise<PutItemSuccessAction | AddItemErrorAction> => {
+      oldText = oldText || getState().list.items.get(id).text;
```
PetrSvirak: While it is definitely funny to pass string parameters around, it gets a bit weird if they are unused 🤣

src/components/App.tsx
```diff
@@ -1,24 +1,48 @@
...
+        </div>
+        <Content />
+      </div>
+      {
+        error &&
+          <ErrorMessage
+            classNames="alert alert-danger error-message"
+            error={error}
+          />
+      }
```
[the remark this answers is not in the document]
medvedovic: @PetrSvirak  You misplaced the closing paren 😛 
```
 {error && (
  <ErrorMessage
    classNames="alert alert-danger error-message"
    error={error}
  />
)}
```
#OCD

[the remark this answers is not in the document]
PetrSvirak: 😱

src/components/ItemError.tsx
```diff
@@ -0,0 +1,56 @@
...
+    <div className="list-item">
+      <ErrorMessage
+        classNames="badge progress-bar-danger item-error"
+        error="Error"
+        tooltip={error}
+      />
+      <span
+        className="glyphicon glyphicon-repeat"
+        onClick={retry}
+        tabIndex={-1}
```
PetrSvirak: I kind of miss tooltip here - how do I know what the icon does? (glyphs are just as google people understand them and if your font doesn't load, they will only see a rectangle here, hence the problem with missing tooltip 😉)

PetrSvirak: BTW: There the tag is a `span`, why do you disable `tabIndex` on it?

src/components/NewListItem.tsx
```diff
@@ -13,6 +13,7 @@ type NewListItemProps = NewListItemDispatchProps;
 
 type NewListItemState = {
   readonly inputText: string,
+  readonly isInputChanged: boolean,
```
PetrSvirak: That's a valid solution (in the sense of https://github.com/KenticoAcademy/kentico-onboarding-js/pull/161#discussion_r266339221), however, you did not apply this philosophy to the `EditListItem` that has the similar problem and looks unnatural now.

